PR #28, opened 2022-02-10
Integrate POC test vectors

dconnolly:


dconnolly:
TODO(dconnolly): move test vector test to separate PR dependent on this one

dconnolly:
> TODO(dconnolly): move test vector test to separate PR dependent on this one

Fixed anyway!

dconnolly:
@chelseakomlo this is now compat with the Ristretto255-SHA512 ciphersuite test vectors

frost-ristretto255/src/frost.rs
```diff
@@ -361,56 +547,220 @@ impl SigningNonces {
...
+impl From<SignatureResponse> for [u8; 64] {
+    fn from(sig: SignatureResponse) -> [u8; 64] {
+        let mut bytes = [0; 64];
+        bytes[0..32].copy_from_slice(&sig.R_share.compress().to_bytes());
+        bytes[32..64].copy_from_slice(&sig.z_share.to_bytes());
+        bytes
+    }
+}
+
+// impl FromHex for SignatureResponse {
```
chelseakomlo, line 721: Remove this commented out code

frost-ristretto255/src/frost.rs
```diff
@@ -655,91 +918,25 @@ pub fn aggregate(
...
     // The aggregation of the signature shares by summing them up, resulting in
     // a plain Schnorr signature.
     let mut z = Scalar::zero();
+    let mut R: RistrettoPoint = RistrettoPoint::identity();
+
     for signature_share in signing_shares {
-        z += signature_share.signature.0;
+        z += signature_share.signature.z_share;
+        // TODO(dconnolly): enforce this == group_commitment? This is not in the spec.
+        R += signature_share.signature.R_share;
```
chelseakomlo, line 935: When deriving the commitment, we should be using values sent strictly in round one. Otherwise, the signer can substitute whatever they want in round two.

chelseakomlo (changes requested):
Everything looks good but in aggregate, we should be generating commitments using values strictly sent in round one.

frost-ristretto255/src/frost.rs
```diff
@@ -635,18 +909,7 @@ pub fn aggregate(
...
-        HashMap::with_capacity(signing_package.signing_commitments.len());
-
-    let rho = generate_rho(signing_package);
-
-    for comm in signing_package.signing_commitments.iter() {
-        bindings.insert(comm.index, rho);
-    }
-
-    let group_commitment = generate_group_commitment(signing_package, &bindings)?;
+    let group_commitment = GroupCommitment::try_from(signing_package)?;
```
chelseakomlo, line 912: Specifically, this line needs to be changed to instead use values sent strictly in round one.

frost-ristretto255/src/frost.rs
```diff
@@ -576,43 +856,37 @@ fn generate_lagrange_coeff(
...
+    let rho: Rho = signing_package.into();
 
-    for comm in signing_package.signing_commitments.iter() {
-        bindings.insert(comm.index, rho);
-    }
-
-    let lambda_i = generate_lagrange_coeff(share_package.index, signing_package)?;
-
-    let group_commitment = generate_group_commitment(signing_package, &bindings)?;
+    let group_commitment = GroupCommitment::try_from(signing_package)?;
```
chelseakomlo, line 865: Similarly, this line needs to be changed to use values strictly sent from round one.

frost-ristretto255/src/frost.rs
```diff
@@ -655,91 +918,25 @@ pub fn aggregate(
...
-            .signing_commitments
-            .iter()
-            .find(|comm| comm.index == signing_share.index)
-            .ok_or("No matching signing commitment for signer")?;
-
-        let commitment_i = signer_commitment.hiding.0
-            + (signer_commitment.binding.0 * bindings[&signing_share.index]);
 
-        signing_share.check_is_valid(&signer_pubkey, lambda_i, commitment_i, challenge)?;
+        signing_share.verify(signer_pubkey, lambda_i, challenge)?;
```
chelseakomlo, line 924: This is incomptable with `verify_signature_share` as defined in the spec.

chelseakomlo (changes requested):
Needs to add compatibility with `verify_signature_share`

frost-ristretto255/src/frost.rs
dconnolly: That is basically what the `SigningPackage` is, the commitments sent via the Coordinator to all parties in round 1, but I would like to make this even more explicit with `round1::SigningCommitments` and other types, in a following PR

frost-ristretto255/src/frost.rs
dconnolly: Because the group commitment is not passed in/calc'd from Round 1 commitments? I'll fix

frost-ristretto255/src/frost.rs
dconnolly: Updated in fadb6fe

dconnolly:
Made some changes @chelseakomlo

frost-ristretto255/src/frost.rs
dconnolly: Updated

frost-ristretto255/src/frost.rs
dconnolly: Updated

frost-ristretto255/src/frost.rs
```diff
@@ -91,27 +269,82 @@ pub(crate) struct Commitment(pub(crate) RistrettoPoint);
...
+        let rho: Rho = signing_package.into();
+
+        let identity = RistrettoPoint::identity();
+        let mut accumulator = identity;
+
+        // Ala the sorting of B, just always sort by index in ascending order
+        //
+        // https://github.com/cfrg/draft-irtf-cfrg-frost/blob/master/draft-irtf-cfrg-frost.md#encoding-operations-dep-encoding
+        for commitment in signing_package.signing_commitments() {
+            // The following check prevents a party from accidentally revealing their share.
```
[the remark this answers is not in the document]
dconnolly: I think this is, if a nonce commitment results in the identity element, and you include it in accumulating the group commitment anyway, that can be used to figure out the nonce itself, or something like that. tl;dr we don't want to compute over commitments that are the identity

frost-ristretto255/src/frost.rs
```diff
@@ -91,27 +269,82 @@ pub(crate) struct Commitment(pub(crate) RistrettoPoint);
...
+            // Note that the '&&' operator would be sufficient.
+            if identity == commitment.binding.0 || identity == commitment.hiding.0 {
```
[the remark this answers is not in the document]
dconnolly: Oh ok, hm. I can update/remove these comments.

frost-ristretto255/src/frost.rs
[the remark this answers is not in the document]
chelseakomlo: The product of two public elements, even one that is the identity, does not result in revealing the nonce. It is true that we should check for the identity, but at worst, it reveals in a key recovery attack for the individual that produced the commitment equal to the identity (so the adversary).

chelseakomlo (approved):

PR #31, opened 2022-03-04
Move keygen, key types, into frost::keys

dconnolly:
Eventually may only export the external API types at the `frost::{}` level but starting here


frost-ristretto255/src/frost/keys.rs
```diff
@@ -0,0 +1,344 @@
...
+/// themselves are scalars) for a secret polynomial f, where f is used to
+/// generate each ith participant's key share f(i). Participants use this set of
+/// commitments to perform verifiable secret sharing.
+///
+/// Note that participants MUST be assured that they have the *same*
+/// [`ShareCommitment`], either by performing pairwise comparison, or by using
+/// some agreed-upon public location for publication, where each participant can
+/// ensure that they received the correct (and same) value.
+#[derive(Clone)]
+pub struct ShareCommitment(pub(crate) Vec<CoefficientCommitment>);
```
chelseakomlo, line 100: A better name for this might be a VSSCommitment- it isn't really committing to a share, it is committing to the polynomial for which that share is a point on. So if the share is `f(i)`, the VSSCommitment is a vector of the coefficients of `f` "in the exponent", so that the verification step is just evaluating `f(i)` (again in the exponent) to make sure it matches the received share.

chelseakomlo (approved):
Looks fantastic! One small rename suggestion.

frost-ristretto255/src/frost/keys.rs
dconnolly: Tried a similar rename: 046d797

chelseakomlo (approved):

PR #52, opened 2022-04-08
Add README.md for frost-core and update for root of repo

dconnolly:
Top level is ok (example will be updated when `frost-ristretto255` is updated to use `frost-core`), `frost-core` will get fleshed out more as the rest of the types are generic-ized over the `Ciphersuite`s


chelseakomlo (approved):
Looks great, although when you merge in the `frost-core` refactor make sure to fill in the example.

PR #73, opened 2022-06-20
 Support naive batch verification

dconnolly:
We have supported this in the earlier implementations, make this generic over `Ciphersuite`s 

This will not be the most optimized for every curve/group, but will be generally fast batch verification.

Generic batch verification compiles, but the batch verify test in core fails :sob:


dconnolly:
@conradoplg Generic batch verification compiles, but the batch verify test in core fails :sob:, if you spot anything obvious

frost-core/src/verifying_key.rs
```diff
@@ -37,12 +37,22 @@ where
     pub fn verify(&self, msg: &[u8], signature: &Signature<C>) -> Result<(), Error> {
         let c = crate::challenge::<C>(&signature.R, &self.element, msg);
 
+        self.verify_prehashed(signature, c)
+    }
+
+    /// Verify a purported `signature` with a pre-hashed [`Challenge`] made by this verification
+    /// key.
+    pub fn verify_prehashed(
```
conradoplg, line 45: Does this need to be public (instead of `pub(crate`)? If yes, I'd suggest using some other name (e.g. `verify_challenge`). "prehashed" is a term used in RFC 8032 to signify something similar but different (passing a hash as the message) and may be confusing for users that are familiar with that RFC

frost-core/src/batch.rs
```diff
@@ -115,49 +110,49 @@ where
...
-            let P_coeff = z * s;
-            P_coeff_acc -= P_coeff;
-
-            R_coeffs.push(z);
+            R_coeffs.push(blind);
             Rs.push(R);
 
-            VK_coeffs.push(Scalar::zero() + (z * c));
-            VKs.push(VK);
+            VK_coeffs.push(<<C::Group as Group>::Field as Field>::zero() + (z * item.c.0));
```
conradoplg, line 127: As discussed in chat

```suggestion
            VK_coeffs.push(<<C::Group as Group>::Field as Field>::zero() + (blind * item.c.0));
```

frost-core/src/verifying_key.rs
dconnolly: I'll rename, even if it's only used in `batch`

dconnolly: Made `pub(crate)` and keeping name for now

conradoplg (approved):
Looks good!

PR #76, opened 2022-06-24
Fix for P-256 support due to identity encoding; other improvements

conradoplg:
While doing the P-256 implementation I had to do a small fix and also noticed some possible simplifications (these are all is separate commits so feel free to reject some of them if needed)

- The `AsMut` bound was not needed (and got in the way trying to make `Serialization` with `p256::Encoding`)
- When decoding a signature, it got the serialized length of the first element by serializing the identity. However in P-256 encoding that is a single byte (as an exception) which made the function fail. I changed it to use the generator instead.
- The `FromHex` bound was not strictly needed, since `TryFrom<Vec<u8>>` is available - we can convert from hex to Vec and them use that trait.
- The `Default` bound for `Group` was not being used. I removed it because it will return the identity, which may cause problems in the P-256 case due to its weird encoding. We can leave it though.

### DIFF
--- a/frost-core/src/signature.rs
+++ b/frost-core/src/signature.rs
@@ -28,7 +28,11 @@ where
         <<C::Group as Group>::Serialization as TryFrom<Vec<u8>>>::Error: Debug,
         <<<C::Group as Group>::Field as Field>::Serialization as TryFrom<Vec<u8>>>::Error: Debug,
     {
-        let mut R_bytes = Vec::from(<C::Group as Group>::Serialization::default().as_ref());
+        // To compute the expected length of the encoded point, encode the generator
+        // and get its length. Note that we can't use the identity because it can be encoded
+        // shorter in some cases (e.g. P-256, which uses SEC1 encoding).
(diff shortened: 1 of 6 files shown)


frost-core/src/signature.rs
```diff
@@ -28,7 +28,11 @@ where
         <<C::Group as Group>::Serialization as TryFrom<Vec<u8>>>::Error: Debug,
         <<<C::Group as Group>::Field as Field>::Serialization as TryFrom<Vec<u8>>>::Error: Debug,
     {
-        let mut R_bytes = Vec::from(<C::Group as Group>::Serialization::default().as_ref());
+        // To compute the expected length of the encoded point, encode the generator
+        // and get its length. Note that we can't use the identity because it can be encoded
+        // shorter in some cases (e.g. P-256, which uses SEC1 encoding).
+        let generator = <C::Group as Group>::generator();
+        let mut R_bytes = Vec::from(<C::Group as Group>::serialize(&generator).as_ref());
```
dconnolly, line 35: Nice, clever

PR #77, opened 2022-06-27
Add P-256 support

conradoplg:
Adds P-256 support. Some notes:

- I didn't include a benchmark because they seem to be currently broken
- My first approach to represent the point encoding was `p256::Encoding`, which handles automatically the annoying identity encoding which takes a single byte. However that ends up causing other problems. I switched to a fixed array in 3c13b9cf. (I realized just after that V5 of the spec does specify that it should be a fixed-length array, so that's better anyway). The only uncertain thing is how the encode the identity, as we discussed in chat. For now I'm just padding it with zeroes.
- There's a lot of repetition in the tests, it's almost everything just a copy-and-paste of ristretto. I wonder if we should make them generic over the ciphersuite?

### DIFF
--- a/frost-p256/README.md
+++ b/frost-p256/README.md
@@ -1,0 +1,32 @@
+An implementation of Schnorr signatures on the P-256 curve for both single and threshold numbers
+of signers (FROST).
+
+## Examples
+
+Creating a `Signature` with a single signer, serializing and deserializing it, and verifying the
+signature:
+
+```rust
+use rand::thread_rng;
+use frost_p256::*;
+
+let msg = b"Hello!";
+
+// Generate a secret key and sign the message
+let sk = SigningKey::new(thread_rng());
+let sig = sk.sign(thread_rng(), msg);
+
+// Types can be converted to raw byte arrays using `from_bytes`/`to_bytes`
+let sig_bytes = sig.to_bytes();
+let pk_bytes = VerifyingKey::from(&sk).to_bytes();
+
+// Deserialize and verify the signature.
+let sig = Signature::from_bytes(sig_bytes)?;
+
+assert!(
+    VerifyingKey::from_bytes(pk_bytes)
+        .and_then(|pk| pk.verify(msg, &sig))
+        .is_ok()
+);
+# Ok::<(), Error>(())
+```
--- a/frost-p256/src/lib.rs
+++ b/frost-p256/src/lib.rs
@@ -38,7 +38,7 @@ impl Field for P256ScalarField {
     }
 
     fn invert(scalar: &Self::Scalar) -> Result<Self::Scalar, Error> {
-        // [`curve25519_dalek::scalar::Scalar`]'s Eq/PartialEq does a constant-time comparison using
+        // [`p256::Scalar`]'s Eq/PartialEq does a constant-time comparison using
         // `ConstantTimeEq`
         if *scalar == <Self as Field>::zero() {
             Err(Error::InvalidZeroScalar)
@@ -88,7 +88,7 @@ impl Group for P256Group {
     /// (1-byte prefix and 32 bytes for the coordinate).
     ///
     /// Note that, in the P-256 spec, the identity is encoded as a single null byte;
-    /// but here we pad with zeroes. This is acceptable as the identity _should_ never 
+    /// but here we pad with zeroes. This is acceptable as the identity _should_ never
     /// be serialized in FROST, else we error.
     ///
     /// [1]: https://secg.org/sec1-v2.pdf
--- a/frost-p256/src/lib.rs
+++ b/frost-p256/src/lib.rs
@@ -2,9 +2,9 @@
 #![deny(missing_docs)]
 #![doc = include_str!("../README.md")]
 
-use p256::elliptic_curve::hash2curve::{hash_to_field, ExpandMsgXmd};
 use p256::{
     elliptic_curve::{
+        hash2curve::{hash_to_field, ExpandMsgXmd},
         sec1::{FromEncodedPoint, ToEncodedPoint},
         Field as FFField, PrimeField,
     },
@@ -117,8 +117,11 @@ impl Group for P256Group {
         let serialized = serialized_point.as_bytes();
         // Sanity check; either it takes all bytes or a single byte (identity).
         assert!(serialized.len() == fixed_serialized.len() || serialized.len() == 1);
+        // Copy to the left of the buffer (i.e. pad the identity with zeroes).
+        // TODO: Note that identity elements shouldn't be serialized in FROST. This will likely become
+        // part of the API and when that happens, we should return an error instead of
+        // doing this padding.
         {
-            // Copy to the left of the buffer (i.e. pad the identity with zeroes).
             let (left, _right) = fixed_serialized.split_at_mut(serialized.len());
             left.copy_from_slice(serialized);
         }
@@ -158,7 +161,8 @@ impl Ciphersuite for P256Sha256 {
     fn H1(m: &[u8]) -> <<Self::Group as Group>::Field as Field>::Scalar {
         let mut u = [P256ScalarField::zero()];
         let dst = CONTEXT_STRING.to_owned() + "rho";
-        hash_to_field::<ExpandMsgXmd<Sha256>, Scalar>(&vec![m], dst.as_bytes(), &mut u).unwrap();
+        hash_to_field::<ExpandMsgXmd<Sha256>, Scalar>(&vec![m], dst.as_bytes(), &mut u)
+            .expect("should never return error according to error cases described in ExpandMsgXmd");
         u[0]
     }
 
@@ -168,7 +172,8 @@ impl Ciphersuite for P256Sha256 {
     fn H2(m: &[u8]) -> <<Self::Group as Group>::Field as Field>::Scalar {
         let mut u = [P256ScalarField::zero()];
         let dst = CONTEXT_STRING.to_owned() + "chal";
-        hash_to_field::<ExpandMsgXmd<Sha256>, Scalar>(&vec![m], dst.as_bytes(), &mut u).unwrap();
+        hash_to_field::<ExpandMsgXmd<Sha256>, Scalar>(&vec![m], dst.as_bytes(), &mut u)
+            .expect("should never return error according to error cases described in ExpandMsgXmd");
         u[0]
     }
 
--- a/frost-p256/src/lib.rs
+++ b/frost-p256/src/lib.rs
@@ -76,49 +76,6 @@ impl Field for P256ScalarField {
     }
 }
 
-/// Wrapper for [`p256::EncodedPoint`].
-#[derive(Default)]
-pub struct P256Serialization(p256::EncodedPoint);
-
-impl AsRef<p256::EncodedPoint> for P256Serialization {
-    fn as_ref(&self) -> &p256::EncodedPoint {
-        &self.0
-    }
-}
-
-impl From<p256::EncodedPoint> for P256Serialization {
-    fn from(encoded_point: p256::EncodedPoint) -> Self {
-        Self(encoded_point)
-    }
-}
-
-impl FromHex for P256Serialization {
-    type Error = &'static str;
-
-    fn from_hex<T: AsRef<[u8]>>(hex: T) -> Result<Self, Self::Error> {
-        let data = hex::decode(hex).map_err(|_| "hex decode error")?;
-        Ok(p256::EncodedPoint::from_bytes(data)
-            .map_err(|_| "point decode error")?
-            .into())
-    }
-}
-
-impl AsRef<[u8]> for P256Serialization {
-    fn as_ref(&self) -> &[u8] {
-        self.0.as_bytes()
-    }
-}
-
-impl TryFrom<Vec<u8>> for P256Serialization {
-    type Error = &'static str;
-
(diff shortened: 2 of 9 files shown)


frost-p256/README.md
```diff
@@ -0,0 +1,32 @@
...
+// Deserialize and verify the signature.
+let sig = Signature::from_bytes(sig_bytes)?;
+
+assert!(
+    VerifyingKey::from_bytes(pk_bytes)
+        .and_then(|pk| pk.verify(msg, &sig))
+        .is_ok()
+);
+# Ok::<(), Error>(())
+```
```
dconnolly, line 32: We should discuss adding a FROST example (on top of the singleton one?) for this and for `-ristretto255`

frost-p256/src/lib.rs
```diff
@@ -0,0 +1,275 @@
+#![allow(non_snake_case)]
+#![deny(missing_docs)]
+#![doc = include_str!("../README.md")]
+
+use elliptic_curve::hash2curve::{hash_to_field, ExpandMsgXmd};
+use hex::FromHex;
```
dconnolly, line 6:
```suggestion
```

frost-p256/src/lib.rs
```diff
@@ -0,0 +1,275 @@
...
+use frost_core::{frost, Ciphersuite, Field, Group};
+
+#[cfg(test)]
+mod tests;
+
+pub use frost_core::Error;
+
+#[derive(Clone, Copy)]
+/// An implementation of the FROST ciphersuite scalar field.
+pub struct P256ScalarField;
```
dconnolly, line 26:
```suggestion
/// An implementation of the FROST P-256 SHA-256 ciphersuite scalar field.
pub struct P256ScalarField;
```

frost-p256/src/lib.rs
```diff
@@ -0,0 +1,275 @@
+#![allow(non_snake_case)]
+#![deny(missing_docs)]
+#![doc = include_str!("../README.md")]
+
+use elliptic_curve::hash2curve::{hash_to_field, ExpandMsgXmd};
+use hex::FromHex;
+use p256::{
+    elliptic_curve::{
```
dconnolly, line 8: Do we need to pull in `elliptic-curve` in the crate dependencies if this is exported by `p256`?

frost-p256/src/lib.rs
```diff
@@ -0,0 +1,275 @@
...
+        Scalar::ZERO
+    }
+
+    fn one() -> Self::Scalar {
+        Scalar::ONE
+    }
+
+    fn invert(scalar: &Self::Scalar) -> Result<Self::Scalar, Error> {
+        // [`curve25519_dalek::scalar::Scalar`]'s Eq/PartialEq does a constant-time comparison using
+        // `ConstantTimeEq`
```
dconnolly, line 43: We should update this comment after investigating what `p256::Scalar` does under the hood for `Eq`, `PartialEq` for comparison (pretty sure it does the right thing just making sure)

frost-p256/src/lib.rs
```diff
@@ -0,0 +1,275 @@
...
+        let field_bytes: &p256::FieldBytes = buf.into();
+        match Scalar::from_repr(*field_bytes).into() {
+            Some(s) => Ok(s),
+            None => Err(Error::MalformedScalar),
+        }
+    }
+}
+
+#[derive(Clone, Copy, PartialEq)]
+/// An implementation of the FROST ciphersuite group.
```
dconnolly, line 80:
```suggestion
/// An implementation of the FROST P-256 ciphersuite group.
```

frost-p256/src/lib.rs
```diff
@@ -0,0 +1,275 @@
...
+
+    type Element = ProjectivePoint;
+
+    /// [SEC 1][1] serialization of a compressed point in P-256 takes 33 bytes
+    /// (1-byte prefix and 32 bytes for the coordinate).
+    /// Note that, in the spec, the identity is encoded as a single null byte;
+    /// bute here we pad with zeroes.
+    ///
+    /// [1]: https://secg.org/sec1-v2.pdf
+    type Serialization = [u8; 33];
```
dconnolly, line 94: Smol tear for 33 bytes instead of 32 😭

frost-p256/src/lib.rs
conradoplg: Good point, removed in fa4f4242c01b6e5031dfa1f6e188ec4604cfab64

frost-p256/src/lib.rs
conradoplg: Oops, done in 2fb7adc. I double checked, it does the same.

Would it make sense to add a trait bound to `ConstantTimeEq` in `frost-core`?

frost-p256/src/lib.rs
```diff
@@ -0,0 +1,275 @@
...
+
+    fn random<R: RngCore + CryptoRng>(rng: &mut R) -> Self::Scalar {
+        Scalar::random(rng)
+    }
+
+    fn random_nonzero<R: RngCore + CryptoRng>(rng: &mut R) -> Self::Scalar {
+        loop {
+            let scalar = Scalar::random(&mut *rng);
+
+            // This impl of `Eq` calls to `ConstantTimeEq` under the hood
```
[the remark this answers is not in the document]
conradoplg: It does, I double checked it

frost-p256/src/lib.rs
conradoplg: :smiling_face_with_tear:

frost-p256/src/lib.rs
```diff
@@ -0,0 +1,275 @@
...
+            .chain("digest")
+            .chain(m);
+
+        let mut output = [0u8; 32];
+        output.copy_from_slice(h.finalize().as_slice());
+        output
+    }
+}
+
+type R = P256Sha256;
```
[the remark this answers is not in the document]
conradoplg: Oops, too much copy&pasting

Changed in 26b8089

frost-p256/README.md
conradoplg: Yep, it's kinda confusing as is. I think we can wait until V5 is ready

frost-p256/src/lib.rs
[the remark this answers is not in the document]
dconnolly: Awesome, thank you!

frost-p256/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+            let (left, _right) = fixed_serialized.split_at_mut(serialized.len());
+            left.copy_from_slice(serialized);
+        }
+        fixed_serialized
+    }
+
+    fn deserialize(buf: &Self::Serialization) -> Result<Self::Element, Error> {
+        let encoded_point =
+            p256::EncodedPoint::from_bytes(buf).map_err(|_| Error::MalformedElement)?;
+        match Option::<AffinePoint>::from(AffinePoint::from_encoded_point(&encoded_point)) {
```
dconnolly, line 131: style nit 

```suggestion

        match Option::<AffinePoint>::from(AffinePoint::from_encoded_point(&encoded_point)) {
```

frost-p256/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+        let serialized = serialized_point.as_bytes();
+        // Sanity check; either it takes all bytes or a single byte (identity).
+        assert!(serialized.len() == fixed_serialized.len() || serialized.len() == 1);
+        {
+            // Copy to the left of the buffer (i.e. pad the identity with zeroes).
+            let (left, _right) = fixed_serialized.split_at_mut(serialized.len());
+            left.copy_from_slice(serialized);
+        }
+        fixed_serialized
+    }
```
dconnolly, line 126: The API for `serialize()` currently always ensures success (as per the FROST spec), but, we have been discussing making the API fallible in case someone tries to serialize() the identity (which should fail with an Error).

This is not in the FROST spec yet and not in the `frost-core` API we're currently implementing, so I'm fine having this as-is but with the forward-looking expectation that `serialize()` will, when updated to implement the applicable version of the FROST spec, become `serialize() -> Result<Self::Serialization, Error>`, similar to `deserialize()`, and instead of panic'ing on the failed assert, return `Err(Error)`

frost-p256/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+            Some(point) => Ok(ProjectivePoint::from(point)),
+            None => Err(Error::MalformedElement),
+        }
+    }
+}
+
+/// Context string from the ciphersuite in the [spec]
+///
+/// [spec]: https://www.ietf.org/archive/id/draft-irtf-cfrg-frost-04.txt
+
```
dconnolly, line 141:
```suggestion
```

frost-p256/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+#![doc = include_str!("../README.md")]
+
+use p256::elliptic_curve::hash2curve::{hash_to_field, ExpandMsgXmd};
+use p256::{
+    elliptic_curve::{
+        sec1::{FromEncodedPoint, ToEncodedPoint},
+        Field as FFField, PrimeField,
+    },
+    AffinePoint, ProjectivePoint, Scalar,
+};
```
dconnolly, line 12: Combine in one `use`?

frost-p256/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+
+    type SignatureSerialization = [u8; 65];
+
+    /// H1 for FROST(P-256, SHA-256)
+    ///
+    /// [spec]: https://www.ietf.org/archive/id/draft-irtf-cfrg-frost-04.html#name-frostp-256-sha-256
+    fn H1(m: &[u8]) -> <<Self::Group as Group>::Field as Field>::Scalar {
+        let mut u = [P256ScalarField::zero()];
+        let dst = CONTEXT_STRING.to_owned() + "rho";
+        hash_to_field::<ExpandMsgXmd<Sha256>, Scalar>(&vec![m], dst.as_bytes(), &mut u).unwrap();
```
dconnolly, line 161: Interesting API

frost-p256/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+            .chain("digest")
+            .chain(m);
+
+        let mut output = [0u8; 32];
+        output.copy_from_slice(h.finalize().as_slice());
+        output
+    }
+}
+
+// Shorthand alias for the ciphersuite
```
dconnolly, line 190: 💖

frost-p256/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+pub mod round1 {
+    use super::*;
+    ///
+    pub type SigningNonces = frost::round1::SigningNonces<P>;
+
+    ///
+    pub type SigningCommitments = frost::round1::SigningCommitments<P>;
+
+    ///
+    pub fn preprocess<RNG>(
```
dconnolly, line 226: We should probably rename this to match the spec in `frost-core` soon, on at least expose a function `commit()` that matches the spec that calls `preprocess(1, id, rng)`

dconnolly: Tracked in https://github.com/ZcashFoundation/frost/issues/78

dconnolly:
This is looking great!

> I didn't include a benchmark because they seem to be currently broken

Yeah we need to fix those up

> My first approach to represent the point encoding was p256::Encoding, which handles automatically the annoying identity encoding which takes a single byte. However that ends up causing other problems. I switched to a fixed array in https://github.com/ZcashFoundation/frost/commit/3c13b9cfaa082d3fd38da752c5705bd364c2f0de. (I realized just after that V5 of the spec does specify that it should be a fixed-length array, so that's better anyway). The only uncertain thing is how the encode the identity, as we discussed in chat. For now I'm just padding it with zeroes.

👍

> There's a lot of repetition in the tests, it's almost everything just a copy-and-paste of ristretto. I wonder if we should make them generic over the ciphersuite?

Yeah I'm not sure the best way to do this. We want them executed over the concrete instantiations, can we do that generically and then call them with the concrete ciphersuites in the top level tests?

conradoplg:
> > There's a lot of repetition in the tests, it's almost everything just a copy-and-paste of ristretto. I wonder if we should make them generic over the ciphersuite?
> 
> Yeah I'm not sure the best way to do this. We want them executed over the concrete instantiations, can we do that generically and then call them with the concrete ciphersuites in the top level tests?

It should be possible but we would need to handle the same issue we have in Zebra with proptests that caused us to have a `proptest-impl` feature: if we do add a generic test code in `frost-core`, it will need to be either public or gated on a feature similar to `proptest-impl`, because if it's behind `#[cgf(test)]` it won't work since only the top-level crate is compiled with the `cfg(test)` and not its dependencies.

frost-p256/src/lib.rs
dconnolly: It's not a strict requirement, more like a nice to have. I don't necessarily want to put that constraint on library users, as it's a third-party trait and not a `std` or `core` trait, even if all the nice libraries _we_ plan to use to implement FROST with support `ConstantTimeEq`, like https://docs.rs/k256/latest/k256/struct.Scalar.html

frost-p256/src/lib.rs
dconnolly: I guess we can remove this now
```suggestion
```

dconnolly:
@Mergifyio refresh

Issue #60, opened 2022-06-01
`frost-core`: update for spec version 5 


dconnolly:
v5: https://www.ietf.org/archive/id/draft-irtf-cfrg-frost-05.html

From the spec changelog:
- [ ] Update test vectors to include version string 
  - [ ] Rename THRESHOLD_LIMIT to MIN_SIGNERS (this will be automatically done when updating test vectors; the code doesn't read that field so no further changes are needed)
- [ ] Use non-contiguous signers for the test vectors 
- [ ] Add a function to generate nonces (#182)
~- [ ] Add MUST that all participants have the same view of VSS commitment~
- [ ] Use THRESHOLD_LIMIT instead of t and MAX_SIGNERS instead of n
~- [x] Remove wire format section (#156)~
- [ ] Update group commitment derivation to have a single scalarmul
- [ ] Use RandomNonzeroScalar for single-party Schnorr example (this should not be needed if we use `nonce_generate` instead)

conradoplg:
Done in https://github.com/ZcashFoundation/frost/pull/89

PR #85, opened 2022-07-12
use a single scalarmul in compute_group_commitment per v5 spec

conradoplg:
Part of https://github.com/ZcashFoundation/frost/issues/60

### DIFF
--- a/frost-core/src/frost.rs
+++ b/frost-core/src/frost.rs
@@ -247,7 +247,8 @@ where
 
         let identity = <C::Group as Group>::identity();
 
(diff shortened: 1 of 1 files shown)


dconnolly:
We will do this to hit v5, fully knowing we will change it back in v6/v7 😂

conradoplg:
> We will do this to hit v5, fully knowing we will change it back in v6/v7 joy

Oops, I didn't notice that :rofl:

dconnolly (approved):
Awesome

PR #84, opened 2022-07-12
add commit(), use it instead of preprocess()

conradoplg:
Closes https://github.com/ZcashFoundation/frost/issues/78

Issue #67, opened 2022-06-20
Update ristretto255 to frost-core V5


conradoplg:
Currently the ristretto255 is working with frost-core V4. Update it to V5.

conradoplg:
Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Engineering-Team-5fb24d9264a3e8000e666a9e/issues/ZcashFoundation/frost/67?planning-poker) with ZenHub @dconnolly

conradoplg:
Done in https://github.com/ZcashFoundation/frost/pull/89

PR #87, opened 2022-07-13
Add nonce_generate function, update test vectors for V5

### LINKED ISSUE #60: `frost-core`: update for spec version 5

### LINKED ISSUE #75: frost-core: Remove `order()` from the `Group` trait
The Order() routine in the spec is only 'consumed' by the definition of the hash functions (H4 etc) and routines like DeserializeScalar(), so it's basically not really a part of the public API, we should remove it.

The order is all over the place but in restricting polynomial coordinates and nonce values etc, which are represented fine in our types (`Scalar` etc).

This specifically impacts p256: `Scalar` should only be required to represent anything in 0 <= x < n while `order()` is n. (This will indeed be the case in P-256 since the p256 crate does not support representing values out of the range).

- [ ] Remove `Group::order()` from `frost-core`
- [ ] Bump version?
- Update known dependent implementations 
  - [ ] `frost-ristretto255`
  - [ ] `frost-p256`

conradoplg:
Depends on #84 

Closes #75 (I would need to update the `order()` documentation so I went ahead and removed it)

Closes #60, #67, #80

## Things to double-check / decide

- I'm not sure what to do with `cofactor()`. It's not used anymore in the spec, and it was removed from the verification procedure. But it seems we will need it back again for V7 so it seems we can just leave it. (Currently it's there, and it's still being used in verification)
- `SigningKey::sign` was not changed to use `nonce_generate`. I see these approaches:
  1. Create a `nonce_generate` function, make `Nonce::new` and `SigningKey::sign` use it (currently the logic is in `Nonce::new`). In which module to put it, though?
  1. Move `Nonce` out of `round1` and into the root module or `signing_key` module, change `Nonce::new` to receive a `Scalar` instead of `KeyPackage`, and use it in `SigningKey::sign`
  1. Don't do anything (current implementation)

I'm currently favoring (i). What do you think?

### DIFF
--- a/frost-core/src/frost/round1.rs
+++ b/frost-core/src/frost/round1.rs
@@ -8,7 +8,7 @@ use zeroize::Zeroize;
 
 use crate::{frost, Ciphersuite, Error, Field, Group};
 
-use super::keys::{KeyPackage, Secret};
+use super::keys::Secret;
 
 /// A scalar that is a signing nonce.
 #[derive(Clone, PartialEq, Zeroize)]
@@ -300,7 +300,8 @@ pub(super) fn encode_group_commitments<C: Ciphersuite>(
 // https://github.com/ZcashFoundation/redjubjub/issues/111
 pub fn preprocess<C, R>(
     num_nonces: u8,
-    key_package: &KeyPackage<C>,
+    participant_index: u16,
+    secret: &Secret<C>,
     rng: &mut R,
 ) -> (Vec<SigningNonces<C>>, Vec<SigningCommitments<C>>)
 where
@@ -312,8 +313,8 @@ where
         Vec::with_capacity(num_nonces as usize);
 
     for _ in 0..num_nonces {
-        let nonces = SigningNonces::new(&key_package.secret_share, rng);
-        signing_commitments.push(SigningCommitments::from((key_package.index, &nonces)));
+        let nonces = SigningNonces::new(secret, rng);
+        signing_commitments.push(SigningCommitments::from((participant_index, &nonces)));
         signing_nonces.push(nonces);
     }
 
@@ -329,12 +330,18 @@ where
 ///
 /// [`commit`]: https://www.ietf.org/archive/id/draft-irtf-cfrg-frost-05.html#section-5.1
 pub fn commit<C, R>(
-    key_package: &KeyPackage<C>,
+    participant_index: u16,
+    secret: &Secret<C>,
     rng: &mut R,
-) -> (Vec<SigningNonces<C>>, Vec<SigningCommitments<C>>)
+) -> (SigningNonces<C>, SigningCommitments<C>)
 where
     C: Ciphersuite,
     R: CryptoRng + RngCore,
 {
-    preprocess(1, key_package, rng)
+    let (mut vec_signing_nonces, mut vec_signing_commitments) =
+        preprocess(1, participant_index, secret, rng);
+    (
+        vec_signing_nonces.pop().expect("must have 1 element"),
+        vec_signing_commitments.pop().expect("must have 1 element"),
+    )
 }
--- a/frost-core/src/frost/round1.rs
+++ b/frost-core/src/frost/round1.rs
@@ -43,6 +43,7 @@ where
             .chain(secret_enc.as_ref().iter())
             .cloned()
             .collect();
+
         Self(C::H4(input.as_slice()))
     }
 
--- a/frost-core/src/frost/round1.rs
+++ b/frost-core/src/frost/round1.rs
@@ -8,7 +8,7 @@ use zeroize::Zeroize;
 
 use crate::{frost, Ciphersuite, Error, Field, Group};
 
-use super::keys::KeyPackage;
+use super::keys::{KeyPackage, Secret};
 
 /// A scalar that is a signing nonce.
 #[derive(Clone, PartialEq, Zeroize)]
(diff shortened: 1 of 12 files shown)


frost-core/src/frost/round1.rs
```diff
@@ -16,21 +18,32 @@ impl<C> Nonce<C>
 where
     C: Ciphersuite,
 {
-    /// Generates a new uniformly random signing nonce.
+    /// Generates a new uniformly random signing nonce by sourcing fresh
+    /// randomness and combining with the secret key, to hedge against a bad
+    /// RNG.
```
dconnolly, line 23: 👍

frost-core/src/frost/round1.rs
```diff
@@ -16,21 +18,32 @@ impl<C> Nonce<C>
...
+
+        let secret_enc =
+            <<C::Group as Group>::Field as Field>::serialize(&key_package.secret_share.0);
+
+        let input: Vec<u8> = k_enc
+            .iter()
+            .chain(secret_enc.as_ref().iter())
+            .cloned()
+            .collect();
+        Self(C::H4(input.as_slice()))
```
dconnolly, line 46:
```suggestion

        Self(C::H4(input.as_slice()))
```

frost-core/src/frost/round1.rs
```diff
@@ -16,21 +18,33 @@ impl<C> Nonce<C>
...
+        let secret_enc =
+            <<C::Group as Group>::Field as Field>::serialize(&key_package.secret_share.0);
+
+        let input: Vec<u8> = k_enc
+            .iter()
+            .chain(secret_enc.as_ref().iter())
+            .cloned()
+            .collect();
+
+        Self(C::H4(input.as_slice()))
```
dconnolly, line 47: Nice

frost-core/src/frost/round1.rs
```diff
@@ -16,21 +18,33 @@ impl<C> Nonce<C>
...
     /// Each participant generates signing nonces before performing a signing
     /// operation.
     ///
-    /// An implementation of `RandomNonzeroScalar()` from the [spec].
+    /// An implementation of `nonce_generate(secret)` from the [spec].
     ///
-    /// [spec]: https://www.ietf.org/archive/id/draft-irtf-cfrg-frost-04.html#section-3.1-3.4
-    pub fn random<R>(rng: &mut R) -> Self
+    /// [spec]: https://www.ietf.org/archive/id/draft-irtf-cfrg-frost-05.html#name-nonce-generation
+    pub fn new<R>(key_package: &KeyPackage<C>, rng: &mut R) -> Self
```
dconnolly, line 31: Should we just restrict this type to `Secret<C>`? This is closer to `nonce_generate()` in the spec and a little more flexible in the API.

frost-core/tests/common/ciphersuite.rs
```diff
@@ -158,6 +154,17 @@ impl Ciphersuite for Ristretto255Sha512 {
...
+    fn H4(m: &[u8]) -> <<Self::Group as Group>::Field as Field>::Scalar {
+        let h = Sha512::new()
+            .chain(CONTEXT_STRING.as_bytes())
+            .chain("nonce")
+            .chain(m);
+
+        let mut output = [0u8; 64];
+        output.copy_from_slice(h.finalize().as_slice());
+        <<Self::Group as Group>::Field as Field>::Scalar::from_bytes_mod_order_wide(&output)
+    }
```
dconnolly, line 167: 👍

frost-p256/src/lib.rs
```diff
@@ -190,6 +185,17 @@ impl Ciphersuite for P256Sha256 {
...
+    ///
+    /// [spec]: https://www.ietf.org/archive/id/draft-irtf-cfrg-frost-05.html#section-6.4-2.2.2.4
+    fn H4(m: &[u8]) -> <<Self::Group as Group>::Field as Field>::Scalar {
+        let mut u = [P256ScalarField::zero()];
+        let dst = CONTEXT_STRING.to_owned() + "nonce";
+        hash_to_field::<ExpandMsgXmd<Sha256>, Scalar>(&vec![m], dst.as_bytes(), &mut u)
+            .expect("should never return error according to error cases described in ExpandMsgXmd");
+        u[0]
+    }
 }
```
dconnolly, line 199: 👍

frost-core/src/frost/round1.rs
dconnolly: So I tried to just make this (or others) `Secret<C>` but of course up the callstack we need to sort the `SigningCommitments` by their identifier (index). But this one and a few others can be `Secret<C>` at least.

frost-core/src/frost/round1.rs
```diff
@@ -314,14 +328,14 @@ where
 /// Generates the signing nonces and commitments to be used in the signing
 /// operation.
 ///
-/// [`commit`]: https://www.ietf.org/archive/id/draft-irtf-cfrg-frost-04.html#name-round-one-commitment
+/// [`commit`]: https://www.ietf.org/archive/id/draft-irtf-cfrg-frost-05.html#section-5.1
 pub fn commit<C, R>(
-    participant_index: u16,
+    key_package: &KeyPackage<C>,
```
dconnolly, line 333: I'm tempted to recommend making the items that we use in KeyPackage explicit: `participant_index` (which should be identifier soon), and the `signing_nonces` which can be extracted by methods (and probably should be in case we make those fields private

dconnolly: This is mostly because `KeyPackage` is a software conceit and `commit()` is from the spec so

dconnolly: > You mean, using `participant_index` and `secret` as parameters instead of `KeyPackage`? I was thinking about encapsulating things to avoid having the user think about how to get the index and the secret. But I don't mind either way. Arguably we could get rind of index since it's not used here, but that would require some refactoring since it's used later.

Yeah basically, it's not great but it's more explicit and closer to the spec, and does not pass in extra stuff in the `KeyPackage` the function doesn't need


> Also, I just realized that I should change the method to return a single `SigningNonces`/`SigningCommitments` instead of a vector. What do you think?

Oh yeah good catch, agreed

conradoplg (commented):
Addressed remaining comments

dconnolly (approved):
YAY

PR #96, opened 2022-07-25
use Identifier instead of index

### LINKED ISSUE #55: Rename participant indexes to participant identifiers or IDs
We now have an `Identifier` type that wraps a `Scalar` field element, for polynomial id's, that is much harder to confused than an `index` into any data structure that may also have an index, like an array.

- [ ] Update all usage of participant ID's to be `Identifier` in `frost-core`
- [ ] Check the public API of `frost-core` to make sure all those are using `Identifier` 
- [ ] Check all known consumers of public `frost-core` API to upgrade them to use `Identifier` where needed
- [ ] If any API changes are made, publish releases of those crates

conradoplg:
Requires #88 

Closes #55 

Things I'm not sure about:

- The commented out `test_identifier_to_scalar` test: should I delete it? Should we make `to_scalar()` public?
- Should we add a `identifier_to_scalar()` method to the trait, and add the implementation in this PR as a default implementation?

### DIFF
--- a/frost-core/src/frost.rs
+++ b/frost-core/src/frost.rs
@@ -99,46 +99,28 @@ where
 
 /// Generates the lagrange coefficient for the i'th participant.
 fn derive_lagrange_coeff<C: Ciphersuite>(
-    signer_id: u16,
+    signer_id: &Identifier<C>,
     signing_package: &SigningPackage<C>,
 ) -> Result<<<C::Group as Group>::Field as Field>::Scalar, &'static str> {
-    // This should fail and panic if signer_id_scalar is 0 in the scalar field.
-    let signer_id_scalar = Identifier::<C>::try_from(signer_id).unwrap();
+    let signer_id_scalar = signer_id.to_scalar();
 
     let zero = <<C::Group as Group>::Field as Field>::zero();
-
-    // TODO: This is redundant
-    if signer_id_scalar.0 == zero {
-        return Err("Invalid parameters");
-    }
-
-    if signing_package
-        .signing_commitments()
-        .iter()
-        .any(|commitment| {
-            let commitment_id_scalar = Identifier::<C>::try_from(commitment.index).unwrap();
-
-            *commitment_id_scalar == zero
-        })
-    {
-        return Err("Invalid parameters");
-    }
 
     let mut num = <<C::Group as Group>::Field as Field>::one();
     let mut den = <<C::Group as Group>::Field as Field>::one();
 
-    // Ala the sorting of B, just always sort by index in ascending order
+    // Ala the sorting of B, just always sort by identifier in ascending order
     //
     // https://github.com/cfrg/draft-irtf-cfrg-frost/blob/master/draft-irtf-cfrg-frost.md#encoding-operations-dep-encoding
     for commitment in signing_package.signing_commitments() {
-        if commitment.index == signer_id {
+        if commitment.identifier == *signer_id {
             continue;
         }
 
-        let commitment_id_scalar = Identifier::<C>::try_from(commitment.index).unwrap();
+        let commitment_id_scalar = commitment.identifier.to_scalar();
 
-        num = num * *commitment_id_scalar;
-        den = den * (*commitment_id_scalar - *signer_id_scalar);
+        num = num * commitment_id_scalar;
+        den = den * (commitment_id_scalar - signer_id_scalar);
     }
 
     if den == zero {
@@ -156,7 +138,7 @@ fn derive_lagrange_coeff<C: Ciphersuite>(
 pub struct SigningPackage<C: Ciphersuite> {
     /// The set of commitments participants published in the first round of the
     /// protocol.
-    signing_commitments: HashMap<u16, round1::SigningCommitments<C>>,
+    signing_commitments: HashMap<Identifier<C>, round1::SigningCommitments<C>>,
     /// Message which each participant will sign.
     ///
     /// Each signer should perform protocol-specific verification on the
@@ -170,32 +152,30 @@ where
 {
     /// Create a new `SigingPackage`
     ///
-    /// The `signing_commitments` are sorted by participant `index`.
+    /// The `signing_commitments` are sorted by participant `identifier`.
     pub fn new(
-        mut signing_commitments: Vec<round1::SigningCommitments<C>>,
+        signing_commitments: Vec<round1::SigningCommitments<C>>,
         message: Vec<u8>,
     ) -> SigningPackage<C> {
-        signing_commitments.sort_by_key(|a| a.index);
-
         SigningPackage {
             signing_commitments: signing_commitments
                 .into_iter()
-                .map(|s| (s.index, s))
+                .map(|s| (s.identifier, s))
                 .collect(),
             message,
         }
     }
 
-    /// Get a signing commitment by its participant index.
-    pub fn signing_commitment(&self, index: &u16) -> round1::SigningCommitments<C> {
-        self.signing_commitments[index]
+    /// Get a signing commitment by its participant identifier.
+    pub fn signing_commitment(&self, identifier: &Identifier<C>) -> round1::SigningCommitments<C> {
+        self.signing_commitments[identifier]
     }
 
     /// Get the signing commitments, sorted by the participant indices
     pub fn signing_commitments(&self) -> Vec<round1::SigningCommitments<C>> {
         let mut signing_commitments: Vec<round1::SigningCommitments<C>> =
             self.signing_commitments.values().cloned().collect();
-        signing_commitments.sort_by_key(|a| a.index);
+        signing_commitments.sort_by_key(|a| a.identifier);
         signing_commitments
     }
 
@@ -250,7 +230,7 @@ where
         let mut group_hiding_commitment = <C::Group as Group>::identity();
         let mut group_binding_commitment = <C::Group as Group>::identity();
 
-        // Ala the sorting of B, just always sort by index in ascending order
+        // Ala the sorting of B, just always sort by identifier in ascending order
         //
         // https://github.com/cfrg/draft-irtf-cfrg-frost/blob/master/draft-irtf-cfrg-frost.md#encoding-operations-dep-encoding
         for commitment in signing_package.signing_commitments() {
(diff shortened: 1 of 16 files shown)


frost-core/tests/tests.rs
```diff
@@ -150,3 +150,18 @@ fn check_sign_with_test_vectors() {
...
+//     type R = Ristretto255Sha512;
+
+//     let one = <<<R as Ciphersuite>::Group as Group>::Field as Field>::one();
+//     let mut sum = <<<R as Ciphersuite>::Group as Group>::Field as Field>::one();
+//     for i in 1..0xFFFFu16 {
+//         let identifier: Identifier<R> = i.try_into().unwrap();
+//         assert_eq!(sum, identifier.to_scalar());
+//         sum = sum + one;
+//     }
+// }
```
conradoplg, line 167: Should we delete this, keep it commented, or make `to_scalar()` public?

frost-core/src/frost/identifier.rs
```diff
@@ -14,183 +14,87 @@ use crate::{Ciphersuite, Error, Field, Group, Scalar};
...
+        let one = <<C::Group as Group>::Field as Field>::one();
+        let mut sum = <<C::Group as Group>::Field as Field>::one();
+        let bits = (self.0.to_be_bytes().len() as u32) * 8;
+        for i in (0..(bits - self.0.leading_zeros() - 1)).rev() {
+            sum = sum + sum;
+            if self.0 & (1 << i) != 0 {
+                sum = sum + one;
+            }
+        }
+        sum
```
conradoplg, line 36: I tried for hours to make this as a `From` impl but Rust doesn't let me.

(If you try to write a `From<Identifier<C>> for Scalar<C>`, Rust does not know that `Scalar<C>` can't be eventually `Identifier<C>` since it's a trait bound which could be possibly be satisfied by `Identifier<C>`.... I think? And it conflicts with the no-op blanket impl of `From<T> for T`)

dconnolly:
> Should we add a identifier_to_scalar() method to the trait, and add the implementation in this PR as a default implementation?

I'm uncertain as of yet, I hope we can settle something in the spec first before exposing this as part of the `Ciphertext` API

frost-core/src/frost.rs
```diff
@@ -99,46 +99,28 @@ where
...
-        .signing_commitments()
-        .iter()
-        .any(|commitment| {
-            let commitment_id_scalar = Identifier::<C>::try_from(commitment.index).unwrap();
-
-            *commitment_id_scalar == zero
-        })
-    {
-        return Err("Invalid parameters");
-    }
```
dconnolly, line 125: ?

conradoplg: `Identifier` is guaranteed to be non-zero per construction, so no need to check it. We can keep the check for security in depth though.

conradoplg: ...but that reminded me to remove `pub(crate)` from the underlying index so that we have more assurance that it won't be constructed with zero, done in dc61839

dconnolly:
@mergify rebase

frost-core/src/frost.rs
dconnolly: Nice

conradoplg:
@dconnolly is this good for merging?

frost-core/src/frost/identifier.rs
```diff
@@ -14,183 +14,87 @@ use crate::{Ciphersuite, Error, Field, Group, Scalar};
...
+pub struct Identifier<C>(u16, PhantomData<C>);
 
-impl<C> AsRef<Scalar<C>> for Identifier<C>
+impl<C> Identifier<C>
 where
     C: Ciphersuite,
 {
-    fn as_ref(&self) -> &Scalar<C> {
-        &self.0
+    // Convert the identifier to a Scalar.
```
dconnolly, line 23:
```suggestion
    // Convert the identifier to a Scalar.
    //
    // Ideally this would be a From<Identifier<C>> for Scalar<C> impl, but rustc
    // doesn't like that
```

frost-core/src/frost/identifier.rs
```diff
@@ -14,183 +14,90 @@ use crate::{Ciphersuite, Error, Field, Group, Scalar};
...
+    //
+    // Ideally this would be a From<Identifier<C>> for Scalar<C> impl, but rustc
+    // doesn't like that
+    pub(crate) fn to_scalar(self) -> Scalar<C> {
+        // Classic left-to-right double-and-add algorithm that skips the first bit 1 (since
+        // identifiers are never zero, there is always a bit 1), thus `sum` starts with 1 too.
+        let one = <<C::Group as Group>::Field as Field>::one();
+        let mut sum = <<C::Group as Group>::Field as Field>::one();
+        let bits = (self.0.to_be_bytes().len() as u32) * 8;
```
dconnolly, line 32:
```suggestion

        let bits = (self.0.to_be_bytes().len() as u32) * 8;
```

frost-core/src/frost/identifier.rs
```diff
@@ -14,183 +14,91 @@ use crate::{Ciphersuite, Error, Field, Group, Scalar};
...
 {
-    fn as_ref(&self) -> &Scalar<C> {
-        &self.0
+    // Convert the identifier to a Scalar.
+    //
+    // Ideally this would be a From<Identifier<C>> for Scalar<C> impl, but rustc
+    // doesn't like that
+    pub(crate) fn to_scalar(self) -> Scalar<C> {
+        // Classic left-to-right double-and-add algorithm that skips the first bit 1 (since
+        // identifiers are never zero, there is always a bit 1), thus `sum` starts with 1 too.
```
dconnolly, line 29: Are we sure that we cannot create an instance of `Identifier` with the value of zero? If not sure, should we do a check here? That would require this method to be a `Result<Scalar<C>, Error>` if we do that

frost-core/src/frost/identifier.rs
```diff
@@ -14,183 +14,91 @@ use crate::{Ciphersuite, Error, Field, Group, Scalar};
...
-        let serialization = bytes
-            .try_into()
-            .map_err(|_| Self::Error::MalformedIdentifier)?;
-
-        let scalar = <<C::Group as Group>::Field as Field>::deserialize(&serialization)?;
-
-        // Participant identifiers are public, so this comparison doesn't need to be constant-time.
-        if scalar == <<C::Group as Group>::Field as Field>::zero() {
+        if n == 0 {
             Err(Self::Error::InvalidZeroScalar)
```
dconnolly, line 99: 👍

frost-core/src/frost/identifier.rs
```diff
@@ -14,183 +14,91 @@ use crate::{Ciphersuite, Error, Field, Group, Scalar};
 /// over, corresponding to some x-coordinate for a polynomial f(x) = y.  MUST NOT be zero in the
 /// field, as f(0) = the shared secret.
 #[derive(Copy, Clone)]
-pub struct Identifier<C: Ciphersuite>(pub(crate) Scalar<C>);
+pub struct Identifier<C>(u16, PhantomData<C>);
```
dconnolly, line 17: 😭

frost-core/src/frost/identifier.rs
```diff
@@ -14,183 +14,91 @@ use crate::{Ciphersuite, Error, Field, Group, Scalar};
 /// over, corresponding to some x-coordinate for a polynomial f(x) = y.  MUST NOT be zero in the
 /// field, as f(0) = the shared secret.
 #[derive(Copy, Clone)]
-pub struct Identifier<C: Ciphersuite>(pub(crate) Scalar<C>);
```
dconnolly, line 17: OH. I see. Hm.

dconnolly: We may want an NonZeroScalar style Scalar, which would just be Identifier, that wraps Scalar, and never can be zero or greater than the space of u16, but operates like a Scalar otherwise, so you can just use the Identifier type in artihmetic or whatever without converting it to a Scalar

https://github.com/zcash/orchard/blob/3faab98e9e82618a0f2d887054e9e28b0f7947dd/src/spec.rs#L95

conradoplg: The hard part of doing that it's that there is no clean ciphersuite-generic way to convert a Scalar back to a uint16 (which is needed when serializing the identifier).

Using Identifier as a Scalar I think would require the `Deref` trait and the Rust docs is somewhat adamant about using it only for smart pointers. (See https://doc.rust-lang.org/std/ops/trait.Deref.html and also https://rust-lang.github.io/api-guidelines/predictability.html#only-smart-pointers-implement-deref-and-derefmut-c-deref. But some people disagree though, [see discussion](https://stackoverflow.com/questions/45086595/is-it-considered-a-bad-practice-to-implement-deref-for-newtypes) :rofl:  )

frost-core/src/frost/identifier.rs
conradoplg: The only code that can build an `Identifier` is in `identifier.rs`, and there only the `impl<C> TryFrom<u16> for Identifier<C>` builds it. So we should be fine.

However there is no reason to not follow your suggestion since all callers of `to_scalar` already return `Result`s.

So I added the check in b5dd0ce72d

frost-core/src/frost/identifier.rs
conradoplg: :ghost:

frost-core/src/frost/identifier.rs
dconnolly: > there is no clean ciphersuite-generic way to convert a Scalar back to a uint16 

True but the `u16` is an implementation detail, not a specification detail, and the spec does enforce `(De)serializeScalar`, which the `identifier` is in reality; so I think we get away with (`Try`)`From`'ing `Identifier` to `u16` by treating it as a `Scalar` that we enforce never has a value greater than `u16::MAX`; I'm going to try this out

conradoplg: > True but the `u16` is an implementation detail, not a specification detail, and the spec does enforce `(De)serializeScalar`, which the `identifier` is in reality; so I think we get away with (`Try`)`From`'ing `Identifier` to `u16` by treating it as a `Scalar` that we enforce never has a value greater than `u16::MAX`; I'm going to try this out

I'm not sure about this... the spec states

> An identifier is an integer value associated with a participant, or signer, and is a value in the range [1, MAX_SIGNERS].

which is effectively a u16. Version 8 also uses `Scalar(identifier)` to convert an identifier to a scalar which wouldn't make sense if identifier was already a scalar.

dconnolly: 'not a specification detail' is not exactly correct but our rendering of Identifier is not fully pinned to the spec, if that makes sense

dconnolly: https://github.com/ZcashFoundation/frost/pull/114

dconnolly: > > True but the `u16` is an implementation detail, not a specification detail, and the spec does enforce `(De)serializeScalar`, which the `identifier` is in reality; so I think we get away with (`Try`)`From`'ing `Identifier` to `u16` by treating it as a `Scalar` that we enforce never has a value greater than `u16::MAX`; I'm going to try this out
> 
> I'm not sure about this... the spec states
> 
> > An identifier is an integer value associated with a participant, or signer, and is a value in the range [1, MAX_SIGNERS].
> 
> which is effectively a u16. Version 8 also uses `Scalar(identifier)` to convert an identifier to a scalar which wouldn't make sense if identifier was already a scalar.

`Scalar(id)` is interesting! This might need to be a per-ciphersuite function!

conradoplg: > `Scalar(id)` is interesting! This might need to be a per-ciphersuite function! This does not explicitly deal in endianness buuuuut

We can do that, and use the implementation in this PR as the default implementation (or not)

dconnolly (approved):
Approving this to unblock our chain of PRs, some of this Identifier typing will be reworked as the per-ciphersuite u16->Scalar definitions and explicit representation of identifiers as Scalars is merged into v09 of the spec

PR #103, opened 2022-07-27
Make verification per-ciphersuite

### LINKED ISSUE #91: v6/v7: Make verification a per-ciphersuite functionality
- [ ] frost-core
- [ ] frost-p256
- [ ] frost-ristretto255

Per
https://github.com/cfrg/draft-irtf-cfrg-frost/pull/219/files

conradoplg:
Based on #96 

Closes #91 

As it stands, the RFC can be interpreted as basically requiring cofactored verification (the cofactor curves require it, and the curves without cofactor don't, so we can always just multiply by the factor even if it's 1 and it will cover both cases). Thus the current implementation already follows the RFC. But in case there is the need to override it, this PR:

- Adds `VerifySignature` method to the `Ciphersuite` trait, with a default implementation with cofactored verification
- Changes the existing `VerifyingKey.verify()` to call that

There are many possible designs here, this is just what seemed to make more sense. We could also:

- Do nothing
- Don't provide a default implementation, which would allow us to remove `Group::cofactor()` if we want, but then we'd need to add one to every ciphersuite

### DIFF
--- a/frost-core/src/lib.rs
+++ b/frost-core/src/lib.rs
@@ -24,6 +24,28 @@ pub use signature::Signature;
 pub use signing_key::SigningKey;
 pub use verifying_key::VerifyingKey;
 
+/// Verify a purported `signature` with a pre-hashed [`Challenge`] made by this verification
+/// key.
+pub(crate) fn verify_prehashed<C: Ciphersuite>(
+    challenge: Challenge<C>,
+    signature: &Signature<C>,
+    public_key: &VerifyingKey<C>,
+) -> Result<(), Error> {
+    // Verify check is h * ( - z * B + R  + c * A) == 0
+    //                 h * ( z * B - c * A - R) == 0
+    //
+    // where h is the cofactor
+    let zB = C::Group::generator() * signature.z;
+    let cA = public_key.element * challenge.0;
+    let check = (zB - cA - signature.R) * C::Group::cofactor();
+
+    if check == C::Group::identity() {
+        Ok(())
+    } else {
+        Err(Error::InvalidSignature)
+    }
+}
+
 /// A prime order finite field GF(q) over which all scalar values for our prime order group can be
 /// multiplied are defined.
 ///
@@ -208,19 +230,7 @@ pub trait Ciphersuite: Copy + Clone {
     ) -> Result<(), Error> {
         let c = crate::challenge::<Self>(&signature.R, &public_key.element, msg);
 
-        // Verify check is h * ( - z * B + R  + c * A) == 0
-        //                 h * ( z * B - c * A - R) == 0
-        //
-        // where h is the cofactor
-        let zB = Self::Group::generator() * signature.z;
-        let cA = public_key.element * c.0;
-        let check = (zB - cA - signature.R) * Self::Group::cofactor();
-
-        if check == Self::Group::identity() {
-            Ok(())
-        } else {
-            Err(Error::InvalidSignature)
-        }
+        verify_prehashed(c, signature, public_key)
     }
 }
 
--- a/frost-core/src/lib.rs
+++ b/frost-core/src/lib.rs
@@ -26,28 +26,6 @@ pub use signature::Signature;
 pub use signing_key::SigningKey;
 pub use verifying_key::VerifyingKey;
 
-/// Verify a purported `signature` with a pre-hashed [`Challenge`] made by this verification
-/// key.
-pub(crate) fn verify_prehashed<C: Ciphersuite>(
-    challenge: Challenge<C>,
-    signature: &Signature<C>,
-    public_key: &VerifyingKey<C>,
-) -> Result<(), Error> {
-    // Verify check is h * ( - z * B + R  + c * A) == 0
-    //                 h * ( z * B - c * A - R) == 0
-    //
-    // where h is the cofactor
-    let zB = C::Group::generator() * signature.z;
-    let cA = public_key.element * challenge.0;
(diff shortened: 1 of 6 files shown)


frost-core/src/lib.rs
```diff
@@ -191,6 +191,32 @@ pub trait Ciphersuite: Copy + Clone {
     ///
     /// [H4]: https://github.com/cfrg/draft-irtf-cfrg-frost/blob/master/draft-irtf-cfrg-frost.md#cryptographic-hash
     fn H4(m: &[u8]) -> <<Self::Group as Group>::Field as Field>::Scalar;
+
+    /// Verify a signature for this ciphersuite. The default implementation uses the "cofactored"
+    /// equation (it multiplies by the cofactor returned by [`Group::cofactor()`]).
+    /// You may override this to provide a tailored implementation, but it must also multiply
+    /// by the cofactor to comply with the RFC.
+    fn VerifySignature(
```
dconnolly, line 199: We can match the standard Rust naming for this function, like `generate()`, etc

```suggestion
    fn verify_signature(
```

frost-core/src/verifying_key.rs
```diff
@@ -35,21 +35,7 @@ where
...
-        let zB = C::Group::generator() * signature.z;
-        let cA = self.element * c.0;
-        let check = (zB - cA - signature.R) * C::Group::cofactor();
-
-        if check == C::Group::identity() {
-            Ok(())
-        } else {
-            Err(Error::InvalidSignature)
-        }
+        C::VerifySignature(msg, signature, self)
```
dconnolly, line 38: 👍

frost-core/src/lib.rs
```diff
@@ -191,6 +191,32 @@ pub trait Ciphersuite: Copy + Clone {
     ///
     /// [H4]: https://github.com/cfrg/draft-irtf-cfrg-frost/blob/master/draft-irtf-cfrg-frost.md#cryptographic-hash
     fn H4(m: &[u8]) -> <<Self::Group as Group>::Field as Field>::Scalar;
+
+    /// Verify a signature for this ciphersuite. The default implementation uses the "cofactored"
+    /// equation (it multiplies by the cofactor returned by [`Group::cofactor()`]).
+    /// You may override this to provide a tailored implementation, but it must also multiply
+    /// by the cofactor to comply with the RFC.
```
dconnolly, line 198:
```suggestion
    /// You may override this to provide a tailored implementation, but if the ciphersuite defines it, it must also multiply
    /// by the cofactor to comply with the RFC.
```

dconnolly (commented):
A few nits, looks great 👍

frost-core/src/lib.rs
conradoplg: done in 27a5b93

frost-core/src/lib.rs
conradoplg: done in 27a5b93

frost-core/src/lib.rs
```diff
@@ -191,6 +191,32 @@ pub trait Ciphersuite: Copy + Clone {
     ///
     /// [H4]: https://github.com/cfrg/draft-irtf-cfrg-frost/blob/master/draft-irtf-cfrg-frost.md#cryptographic-hash
     fn H4(m: &[u8]) -> <<Self::Group as Group>::Field as Field>::Scalar;
+
+    /// Verify a signature for this ciphersuite. The default implementation uses the "cofactored"
+    /// equation (it multiplies by the cofactor returned by [`Group::cofactor()`]).
+    /// You may override this to provide a tailored implementation, but if the ciphersuite defines it,
+    /// it must also multiply by the cofactor to comply with the RFC.
+    fn verify_signature(
```
dconnolly, line 199:
```suggestion
    /// # Cryptographic Safety
    /// 
    /// You may override this to provide a tailored implementation, but if the ciphersuite defines it,
    /// it also multiply by the cofactor to comply with the RFC.
    fn verify_signature(
```

dconnolly (approved):

conradoplg:
I fixed the conflicts, but I'm not sure how to handle `verify_prehashed`. I moved it to a crate function that is now called by the default implementation of `verify_signature`. But this means that if the ciphersuite provides a custom implementation of `verify_signature`, then the batch verifier won't use it.

The alternative would be to add `verify_prehashed` to the ciphersuite. @dconnolly what do you think?

dconnolly:
Aaah good point. Hm. 

Could duplicate this somewhat. 

For now, for first-pass support, we can just have a TODO that basically says "batch verification supports Schnorr-ish verification equations and does the cofactor multiplication, if your concrete impl of Ciphersuite has a verify_signature() that diverges, you may not want to also use the batch math

dconnolly:
> Aaah good point. Hm. 
> 
> Could duplicate this somewhat. 
> 
> For now, for first-pass support, we can just have a TODO that basically says "batch verification supports Schnorr-ish verification equations and does the cofactor multiplication, if your concrete impl of Ciphersuite has a verify_signature() that diverges, you may not want to also use the batch math

So that may mean we put batch support behind feature flag

conradoplg:
> Aaah good point. Hm.
> 
> Could duplicate this somewhat.
> 
> For now, for first-pass support, we can just have a TODO that basically says "batch verification supports Schnorr-ish verification equations and does the cofactor multiplication, if your concrete impl of Ciphersuite has a verify_signature() that diverges, you may not want to also use the batch math

Sounds good, added a comment in a645a5b

dconnolly (approved):

PR #105, opened 2022-07-28
Make tests generic

conradoplg:
Based on #103 

Create generic version of the tests in `frost-core`. Make them available if the `test-impl` feature is set.

Change p256 and ristretto255 implementations to call the generic ones.


frost-core/src/tests/vectors.rs
```diff
@@ -0,0 +1,274 @@
...
+
+/// Test with the given test vectors for a ciphersuite.
+pub fn check_sign_with_test_vectors<C: Ciphersuite + PartialEq>(json_vectors: &Value)
+where
+    <<C::Group as Group>::Serialization as TryFrom<Vec<u8>>>::Error: std::fmt::Debug,
+    <<<C::Group as Group>::Field as Field>::Serialization as TryFrom<Vec<u8>>>::Error:
+        std::fmt::Debug,
+    <<C as Ciphersuite>::SignatureSerialization as std::convert::TryFrom<std::vec::Vec<u8>>>::Error:
+        std::fmt::Debug,
+    C::Group: PartialEq,
```
conradoplg, line 153: I honestly don't know why the `PartialEq` bounds are required (in fact, I also don't know why the `Debug` bounds are required, but I saw them being used elsewhere). I'll study this later, but it works :stuck_out_tongue:

conradoplg: While working on restoring coverage I figured out the `Debug` trait bound issue: `unwrap()` requires that the error type implements `Debug`. We can't currently express that in our `TryFrom` bounds (e.g. `    type Serialization: AsRef<[u8]> + Debug + Default + TryFrom<Vec<u8>>`) because [associated type bounds](https://github.com/rust-lang/rust/issues/52662) are not supported yet. I switched to using the `debugless-unwrap` and that allows us to remove all these Debug bounds.

I'll look into `PartialEq` next :laughing:

conradoplg (commented):
I restored a serialization test which I accidentally used a commented out version for the generic code (thankfully we have coverage and I noticed through that). I also managed to remove the Debug bounds by using the `debugless-unwrap` crates, see comment.

frost-core/src/tests/vectors.rs
conradoplg: I figured `PartialEq` out [with SO's help](https://stackoverflow.com/questions/73156316/why-partialeq-bound-is-needed-when-comparing-associated-types). We'll need to manually implement `PartialEq` for some types instead of deriving it. I'll do that in a later PR to make it less confusing.

conradoplg: PR here: https://github.com/ZcashFoundation/frost/pull/107

dconnolly:
@mergify rebase

conradoplg:
Conflicts were solved

frost-p256/src/tests.rs
```diff
@@ -1,153 +1,24 @@
...
+use serde_json::Value;
 
 use crate::*;
 
-mod vectors;
-
-use vectors::*;
+lazy_static! {
+    pub static ref P256_SHA256: Value =
+        serde_json::from_str(include_str!("tests/vectors.json").trim())
```
dconnolly, line 9: Wait, where are these p256 vectors?

dconnolly: OH i'm sorry they are in `src/tests/`, not `tests/` as in 'integration' tests for the crate

dconnolly (approved):
🎉

PR #108, opened 2022-07-29
Derive verification share and group public key from commitments

### LINKED ISSUE #47: Note in Ristretto255 near PublicKeyPackage where we impl `derive_group_info()` for v05

conradoplg:
Based on #106 

Closes #47 

This changes `SharePackage` to not contain the group and participant public keys, and instead derives them from the commitments. This was already explictly done in `SecretShare.verify()` and I just changed it to return those values.

### DIFF
--- a/frost-core/src/frost/keys.rs
+++ b/frost-core/src/frost/keys.rs
@@ -248,7 +248,9 @@ where
         &self.value
     }
 
-    /// Verifies that a secret share is consistent with a verifiable secret sharing commitment.
+    /// Verifies that a secret share is consistent with a verifiable secret sharing commitment,
+    /// and returns the derived group info for the participant (their public verification share,
+    /// and the group public key) if successful.
     ///
     /// This ensures that this participant's share has been generated using the same
     /// mechanism as all other signing participants. Note that participants *MUST*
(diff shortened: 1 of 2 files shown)


conradoplg:
@dconnolly rebased and solved conflicts, this is ready for review

frost-core/src/frost/keys.rs
```diff
@@ -275,7 +279,11 @@ where
             return Err("SecretShare is invalid.");
         }
 
-        Ok(())
+        let group_public = VerifyingKey {
+            element: self.commitment.0[0].0,
```
dconnolly, line 283: Hm this API isn't the best (getting the group verifying key out of `VerifiableSecretSharingCommitment`) but whatever

conradoplg: Do you mean conceptually, or this specific code? I agree `self.commitment.0[0].0` is terribly ugly and it's coming up a lot in the DKG code too. Maybe I can add a `impl From<&VerifiableSecretSharingCommitment> for VerifyingKey`?

conradoplg: @dconnolly what do you think?

dconnolly: Ah yes this is just a style nit, we can try to make better in the DKG work perhaps, not a blocker

frost-core/src/frost/keys.rs
```diff
@@ -290,10 +298,6 @@ pub struct SharePackage<C: Ciphersuite> {
     pub identifier: Identifier<C>,
     /// This participant's secret share.
     pub secret_share: SecretShare<C>,
-    /// This participant's public key.
```
dconnolly, line 293: We can address this deduping in a separate PR re this issue: https://github.com/ZcashFoundation/frost/issues/127

dconnolly (approved):

PR #110, opened 2022-08-02
add ristretto255 docs

### LINKED ISSUE #61: Flesh out all rustdoc in `frost-ristretto255` public API
Half of the public items are well-doc'd, half just have placeholders to allow compilation.

conradoplg:
Based on #108 

Closes #61 

I just copied the docs from frost-core with some simplifications. Let me know if that's OK and if it's better to mention Ristretto255 explicitly, or if that's just implied.

After we decide on this I can write a PR for P-256.


dconnolly:
> Based on #108 
> 
> Closes #61 
> 
> I just copied the docs from frost-core with some simplifications. Let me know if that's OK and if it's better to mention Ristretto255 explicitly, or if that's just implied.
> 
> After we decide on this I can write a PR for P-256.

Might want to `s/FROST/FROST(ristretto255, SHA-512)/` everywhere that is appropriate and basically do the same for P-256 ✍️

conradoplg:
> Might want to `s/FROST/FROST(ristretto255, SHA-512)/` everywhere that is appropriate and basically do the same for P-256 writing_hand

Done in df865bef15

frost-ristretto255/src/lib.rs
```diff
@@ -198,28 +199,51 @@ pub mod keys {
         frost::keys::keygen_with_dealer(num_signers, threshold, &mut rng)
     }
 
+    /// Secret and public key material generated by a dealer performing
+    /// [`keygen_with_dealer`].
     ///
+    /// To derive a FROST(ristretto255, SHA-512) keypair, the receiver of the [`SharePackage`] *must* call
+    /// .into(), which under the hood also performs validation.
```
dconnolly, line 206:
```suggestion
    /// # Safety
    ///
    /// To derive a FROST(ristretto255, SHA-512) keypair, the receiver of the [`SharePackage`] *must* call
    /// .into(), which under the hood also performs validation.
```

dconnolly: I don't know if cryptographic/security safety is 'the same' as memory unsafety in the rust doc sense, but it's still about requirements the API user must adhere to:

frost-ristretto255/src/lib.rs
```diff
@@ -198,28 +199,51 @@ pub mod keys {
...
 }
 
-///
+/// FROST(ristretto255, SHA-512) Round 1 functionality and types.
 pub mod round1 {
     use frost_core::frost::keys::SecretShareValue;
 
     use super::*;
+
+    /// Comprised of hiding and binding nonces.
```
dconnolly, line 230:
```suggestion
    /// Comprised of FROST(ristretto255, SHA-512) hiding and binding nonces.
```

frost-ristretto255/src/lib.rs
```diff
@@ -232,20 +256,30 @@ pub mod round1 {
...
+/// each signing party.
 pub type SigningPackage = frost::SigningPackage<R>;
 
-///
+/// FROST(ristretto255, SHA-512) Round 2 functionality and types, for signature share generation.
 pub mod round2 {
     use super::*;
 
-    ///
+    /// A participant's signature share, which the coordinator will aggregate with all other signer's
```
dconnolly, line 267:
```suggestion
    /// A FROST(ristretto255, SHA-512) participant's signature share, which the Coordinator will aggregate with all other signer's
```

frost-ristretto255/src/lib.rs
```diff
@@ -255,10 +289,24 @@ pub mod round2 {
     }
 }
 
-///
+/// A Schnorr signature on FROST(ristretto255, SHA-512).
 pub type Signature = frost_core::Signature<R>;
 
+/// Verifies each participant's signature share, and if all are valid,
```
dconnolly, line 295:
```suggestion
/// Verifies each FROST(ristretto255, SHA-512) participant's signature share, and if all are valid,
```

frost-ristretto255/src/lib.rs
```diff
@@ -255,10 +289,24 @@ pub mod round2 {
...
+/// aggregates the shares into a signature to publish.
 ///
+/// Resulting signature is compatible with verification of a plain SpendAuth
```
dconnolly, line 298:
```suggestion
/// Resulting signature is compatible with verification of a plain Schnorr 
```

frost-ristretto255/src/lib.rs
conradoplg: Maybe "Security" would be more appropriate?

dconnolly (approved):

PR #114, opened 2022-08-22
Refreshed Identifier newtype of Scalar with traits

dconnolly:
@conradoplg thoughts?


frost-core/src/frost/identifier.rs
```diff
@@ -14,73 +14,132 @@ use crate::{Ciphersuite, Error, Field, Group, Scalar};
...
+    C: Ciphersuite,
+{
+    fn from(id: Identifier<C>) -> u16 {
+        let mut bytes = [0u8; 2];
+
+        let serialized = <<C::Group as Group>::Field as Field>::serialize(&id.0);
+
+        bytes.copy_from_slice(&serialized.as_ref()[..2]);
+
+        u16::from_le_bytes(bytes)
```
conradoplg, line 43: Unfortunately this does not always work because there is no guarantee that the scalar is serialized in little endian (which was the motivation for this [PR](https://github.com/cfrg/draft-irtf-cfrg-frost/pull/229))

It's passing CI here because our tests don't use indices larger than 1 byte, but it will fail in that case. I pushed a test vector that demonstrates this in 9a3ffda3bea4aa343bcc2b1a7ccd8093efa661f8

dconnolly: funny nit, encode_u16 is big endian

conradoplg: That depends on how you interpret the spec... if you consider that the identifier is an "integer", then you need `integer -> Scalar` and `integer -> encoded u16`. The first one is doable if a bit awkward and the second is easy.

If you consider that an identifier is a Scalar, then you will still need `integer -> Scalar` because you will need to build the scalar somehow, and  `Scalar -> encoded u16`. And I don't see a ciphersuite-generic way to do the `Scalar -> encoded u16` conversion. What you must guarantee is that the index 1 is mapped to the Scalar 1 and so on, which is not happening here. I feel that this second approach is trying to force a leaky abstraction somehow.

dconnolly: identifiers are created and operated over as scalar field elements during key gen operations over the secret polynomial, so 'at heart' they are scalars that we arbitrarily cap in value at MAX_SIGNERS. I do think the spec is not doing the Scalar <-> uint16 enforcement any favors here, and maybe we need to define a per-ciphersuite routine for this or a generic routine _for internal operations only_ (not deserialization/serialization b/c p256, etc) to make this explicit

conradoplg: I'm not sure. Like I mentioned the spec defines identifier as integers, which are now explicitly converted to scalar with `Scalar(i)` in the spec. I wouldn't mind either way, but I don't see the problem with the current approach of using u16:

- It doesn't require the Scalar -> encoded u16 convertion, which does not seem possible to do in a ciphersuite-generic way without a new trait method
- Seems functionally equal in every other aspect (maybe there is a small performance penalty of converting multiple times, but it seems negligible, and could be improved by storing both u16 and Scalar, if needed)

dconnolly:
@mergify rebase

frost-core/src/frost/identifier.rs
conradoplg: (We decided to go ahead and define identifiers as scalar, including in the spec)

conradoplg:
Synced with main and solved conflicts, this is now ready for review.

To implement `PartialOrd/Ord` for `Identifier`, I considered two approaches:

- Adding the same bounds to Scalar. However this seemed bad because e.g. curve25519 Scalar does not have these bounds so we'd need to newtype it.
- Adding a new trait method `little_endian_serialize` to `Field` and use its result to implement the bounds. It's a bit ugly but it will also allow us to quickly solve #141, so I went for this one.

mpguerra:
@dconnolly can you please review Conrado's latest changes? I'm not able to assign you as a reviewer as this was your issue originally :) 

@conradoplg I think we'll need for you to do the final approval and merge after Deirdre's done

frost-core/src/lib.rs
```diff
@@ -76,6 +74,12 @@ pub trait Field: Copy + Clone {
     /// <https://www.ietf.org/archive/id/draft-irtf-cfrg-frost-10.html#section-3.1-3.8>
     fn serialize(scalar: &Self::Scalar) -> Self::Serialization;
 
+    /// A member function of a [`Field`] that maps a [`Scalar`] to a unique byte array buf of
+    /// fixed length Ne, in little-endian order.
+    ///
+    /// This is used internally.
+    fn little_endian_serialize(scalar: &Self::Scalar) -> Self::Serialization;
```
dconnolly, line 81: Not my favorite that we need this but fine with it

dconnolly (commented):
I can't approve but LGTM!

PR #112, opened 2022-08-04
use per-signer Rho per V7 spec

conradoplg:
Based on #110 

Closes #92

Issue #63, opened 2022-06-17
frost-ristretto255: implement `frost-core`'s DKG concretely


dconnolly:
When https://github.com/ZcashFoundation/frost/issues/35 is done, use that to provide DKG for the Ristretto255 ciphersuite

- [ ] Pull in supporting `frost-core` version
- [ ] Implement for concrete `FrostRistretto255` `Ciphersuite` 
- [ ] Release minor version of `frost-ristretto255` crate

PR #129, opened 2022-09-21
Add DKG support

### LINKED ISSUE #35: frost-core: Implement the generic DKG
Implement the distributed key generation in `frost-core`, generic over the ciphersuite. 

- [ ] Implement generic DKG
- [ ] Release minor version update for `frost-core`

conradoplg:
Implements DKG per the FROST paper.

This required adding a new hash function to the Ciphersuite. I named it `HDKG` instead of `H6` to avoid future collisions. I also made it optional by returning an `Option`, so that ciphersuites can return None if they don't need the DKG, and making returning None the default implementation so that ciphersuites don't even need to implement it if not needed.

I'm not super happy with the struct names, suggestions are welcome. Though I think that using "Round1/2" in them makes it a bit clearer for users than trying to come up with semantic names.

Based on #112 

Closes #35 , #63


conradoplg:
@dconnolly I rebased and solved conflicts. I reproduced the issue you found, it was caused by a mistake in `generate_secret_shares`. My intention was for it to prepend the secret to the coefficient vector and return it, matching the spec. But it was returning the original vector, which caused problems when rebasing it here due to the refactorings.

I changed `generate_secret_shares` to return the full vector in  418500a39e58cd6b6d5e56b9600ae5a6ac529269, and that allowed me to simplify `evaluate_polynomial` in cb657a978b6e2378e58b3d5fd3bb5f9e1283768f

frost-core/src/tests.rs
```diff
@@ -131,3 +140,180 @@ pub fn check_sign_with_dealer<C: Ciphersuite, R: RngCore + CryptoRng>(mut rng: R
...
+    // Map of the verifying key of each participant.
+    // Used by the signing test that follows.
+    let mut verifying_keys = HashMap::new();
+    // The group public key, used by the signing test that follows.
+    let mut group_public = None;
+    // For each participant, store the set of verifying keys of the other
+    // participants. This is used to check if the set is correct for all
+    // participants.
+    // In practice, if there is a Coordinator, only they need to store the set.
+    // If there is not, them all candidates must store their own sets.
```
upbqdn, line 268:
```suggestion
    // If there is not, then all candidates must store their own sets.
```

upbqdn (approved):
The PR looks very good. I tried to thoroughly check if the implementation matches the spec and the paper. I just have a few very minor suggestions.

Note that running
```bash
cargo d --document-private-items
```
produces two warnings about unresolved links.

conradoplg:
> The PR looks very good. I tried to thoroughly check if the implementation matches the spec and the paper. I just have a few very minor suggestions.
> 
> Note that running
> 
> ```shell
> cargo d --document-private-items
> ```
> 
> produces two warnings about unresolved links.

Thanks, great suggestions! I also fixed the warnings.

upbqdn (approved):

PR #130, opened 2022-09-21
Add documentation generator; add frost-p256 documentation using it

### LINKED ISSUE #95: Flesh out all rustdoc in frost-p256 public API
Half of the public items are well-doc'd, half just have placeholders to allow compilation.

conradoplg:
Since all ciphersuites documentation are basically the same, I added a Rust program to generate them from the canonical ristretto255 implementation. Using it, I added the P-256 docs.

Closes #95


dconnolly:
A quick nit before diving in: any chance this could be written in Rust? 🙈

conradoplg:
> A quick nit before diving in: any chance this could be written in Rust? see_no_evil

Why not? :rofl: 
I'll see if I can convert it without spending too much time on it

conradoplg:
Done! :snake: :arrow_right:  :crab:

dconnolly:
> Done! 🐍 ➡️ 🦀

Wow wow wow!

Cargo.toml
```diff
@@ -6,3 +6,7 @@ members = [
     "frost-ristretto255",
     "frost-p256",
 ]
+
+# This is a development helper binary.
+# To run it, execute `cargo run --manifest-path gendoc/Cargo.toml`
+exclude = ["gendoc"]
```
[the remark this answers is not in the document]
dconnolly: We can make sure it doesn't get published while still keeping it maintained as part of the Rust workspace I think

dconnolly, line 12:
```suggestion
    "frost-p256",
    "gendoc"
]
```

gendoc/src/main.rs
```diff
@@ -0,0 +1,123 @@
+//! Generate documentation for a ciphersuite based on another ciphersuite implementation.
+//!
+//! The documentation for each ciphersuite is very similar, with the only difference being
+//! the ciphersuite name.
+//!
+//! To make it easier to update all ciphersuites documentations when a change is needed,
+//! this binary allows updating all of them based on a single one. This binary
+//! uses frost-ristretto255 as the "canonical" one, so:
```
dconnolly, line 8:
```suggestion
//! To make it easier to update all ciphersuite documentation when a change is needed,
//! this binary allows updating all of them based on a single one. This binary
//! uses frost-ristretto255 as the "canonical" one, so:
```

gendoc/src/main.rs
```diff
@@ -0,0 +1,123 @@
...
+//!
+//! - Change any documentation of a public function or struct in `frost-ristretto255/src/lib.rs`
+//! - Run `cargo run --manifest-path gendoc/Cargo.toml` to update the documentation
+//!   of the other ciphersuites.
+
+use std::{fs, iter::zip};
+
+use regex::Regex;
+
+/// Read the public documentation of public symbols in the given file.
```
dconnolly, line 18:
```suggestion
/// Read the public documentation of public types in the given file.
```

dconnolly: I think this is correct?

gendoc/src/main.rs
```diff
@@ -0,0 +1,123 @@
...
+/// suite_names_code: strings that reference the specific suite in code
+///     inside `fn` and should be ignore when using for replacements.
+///
+/// # Returns
+///
+/// A list with data for each symbol, see above.
+fn read_docs(filename: &str, suite_names_code: &[&str]) -> Vec<(String, String, usize, usize)> {
+    let mut docs = Vec::new();
+    let code = fs::read_to_string(filename).unwrap();
+    let re = Regex::new(r"(?m)((^[ ]*///.*\n)+)\s*pub (.*)").unwrap();
```
dconnolly, line 46: I'm fine with regex'ing for now as this is an on-demand internal tool - if we are using it a lot, I wonder if there is some library API to hook into rustdoc processing itself, as the regex may be harder to maintain

Cargo.toml
dconnolly: I won't just apply this as it makes `gendoc` the default executable when `cargo run` is done in the root of the workspace, which I'm not sure we want?

gendoc/src/main.rs
conradoplg: I was looking for a word that covers types, functions, constants...

gendoc/src/main.rs
conradoplg: Some time ago I tried this with `syn` but it's a bit annoying, it's not well documented, and I managed just to extract the docs, actually replacing them would be more difficult... so I went with the simplest thing that worked. But yeah, not a fan of regexes myself :sweat_smile:

Cargo.toml
conradoplg: I'll try tinkering with the manifest to make it work

conradoplg: If we use `default-members` then `cargo run` won't run it, but you will need to use `cargo check --workspace` to check everything. What do you think?

gendoc/src/main.rs
dconnolly: Perhaps 'items': https://doc.rust-lang.org/reference/items.html

conradoplg: Ah neat, thanks!

conradoplg: Replaced in aa95773

Cargo.toml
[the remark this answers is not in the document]
dconnolly: Resolved for now with https://github.com/ZcashFoundation/frost/pull/130/commits/07ebaa3a1608a7acb96209f2ac80f06e264b6af1

dconnolly: Resolved for now with https://github.com/ZcashFoundation/frost/pull/130/commits/07ebaa3a1608a7acb96209f2ac80f06e264b6af1

dconnolly (approved):

dconnolly:
@mergify rebase

dconnolly (dismissed):

dconnolly (approved):

Issue #141, opened 2022-09-30
Change batching to work with big-endian-encoded scalars


conradoplg:
The current implementation assumes Field::serialize will return a little-endian array, which is not always true (e.g. P256).

One possibility is adding a new `serialize_little_endian` method to the `Field` trait and used that.

mpguerra:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Engineering-Team-5fb24d9264a3e8000e666a9e/issues/ZcashFoundation/frost/141?planning-poker) with Zenhub @conradoplg @dconnolly

conradoplg:
> Should this be closed now that #142 is merged or is there more work in this?

There is more work, this was a bug found during that ticket that it's not obvious how to fix so I opened a ticket for it

PR #142, opened 2022-09-30
make batch testing generic; remove duplicated Ristretto255 implementation

conradoplg:
We want to remove the duplicated Ristretto255 implementation that was used in `frost-core`, since we will test it anyway indirectly with the curve-specific tests.

To do that, we need to make the batch testing generic.

During this I found out that the current batch implementation only works for little-endian-encoded scalars, I created https://github.com/ZcashFoundation/frost/issues/141 to track that.


conradoplg:
Rebased

dconnolly (approved):
Love deleting code 🎉

image

PR #145, opened 2022-10-04
reject identity in Group::deserialize

### LINKED ISSUE #134: Reject identity in DeserializeElement
Change ristretto255 and P-256 implementations of DeserializeElement to reject the identity.

In P-256 this will allow some simplifications, refer to comments in the code.

conradoplg:
Closes #134

While doing this I wondered if we should do the same in `serialize`, this is being currently discussed in chat, and we can do that in another PR if we want to.

### DIFF
--- a/frost-p256/src/lib.rs
+++ b/frost-p256/src/lib.rs
@@ -4,6 +4,7 @@
 
 use p256::{
     elliptic_curve::{
+        group::prime::PrimeCurveAffine,
         hash2curve::{hash_to_field, ExpandMsgXmd},
         sec1::{FromEncodedPoint, ToEncodedPoint},
         Field as FFField, PrimeField,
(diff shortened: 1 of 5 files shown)


frost-ristretto255/tests/frost.rs
```diff
@@ -21,3 +23,11 @@ fn check_bad_batch_verify() {
 
     frost_core::tests::batch::bad_batch_verify::<Ristretto255Sha512, _>(rng);
 }
+
+#[test]
+fn check_deserialize_identity() {
+    let encoded_identity = RistrettoPoint::identity().compress().to_bytes();
+
+    let r = <<Ristretto255Sha512 as Ciphersuite>::Group as Group>::deserialize(&encoded_identity);
+    assert_eq!(r, Err(Error::InvalidIdentityElement));
```
dconnolly, line 32: 👍

frost-p256/src/lib.rs
```diff
@@ -128,7 +129,16 @@ impl Group for P256Group {
             p256::EncodedPoint::from_bytes(buf).map_err(|_| Error::MalformedElement)?;
 
         match Option::<AffinePoint>::from(AffinePoint::from_encoded_point(&encoded_point)) {
-            Some(point) => Ok(ProjectivePoint::from(point)),
+            Some(point) => {
+                if point.is_identity().into() {
+                    // This is actually impossible since the identity is encoded a a single byte
+                    // which will never happen since we receive a 33-byte buffer.
+                    // We leave the check for consistency.
+                    Err(Error::InvalidIdentityElement)
```
dconnolly, line 137: 💖

dconnolly (approved):

dconnolly:
> While doing this I wondered if we should do the same in serialize, this is being currently discussed in chat, and we can do that in another PR if we want to.

Discussed elsewhere, we think this should be reflected in a spec change to also reject the identity element when trying to SerializeElement

PR #146, opened 2022-10-04
merge SharePackage into SecretShare

### LINKED ISSUE #127: Dedupe `SharePackage` and `SecretShare`
Once upon a time the original `SecretShare` type was much simpler, but now the `SharePackage` type is basically redundant around what `SecretShare` already has. I think the places that expect / return `SharePackage` can be replaced with `SecretShare`

conradoplg:
Closes #127 

In #35 I created a lot of types with "Package" in their names, inspired by the `SharePackage` type, so we might want to rethink those names :sweat_smile:


frost-core/src/frost/keys.rs
```diff
@@ -233,13 +233,17 @@ pub struct VerifiableSecretSharingCommitment<C: Ciphersuite>(
     pub(super) Vec<CoefficientCommitment<C>>,
 );
 
-/// A secret share generated by performing a (t-out-of-n) secret sharing scheme.
+/// A secret share generated by performing a (t-out-of-n) secret sharing scheme,
+/// generated by a dealer performing [`keygen_with_dealer`].
```
dconnolly, line 237: We'll probably update this when the DKG is merged?

frost-core/src/frost/keys.rs
```diff
@@ -233,13 +233,17 @@ pub struct VerifiableSecretSharingCommitment<C: Ciphersuite>(
...
 ///
 /// `n` is the total number of shares and `t` is the threshold required to reconstruct the secret;
 /// in this case we use Shamir's secret sharing.
 ///
 /// As a solution to the secret polynomial _f_ (a 'point'), the `identifier` is the x-coordinate, and the
 /// `value` is the y-coordinate.
+///
+/// To derive a FROST keypair, the receiver of the [`SecretShare`] *must* call
+/// .into(), which under the hood also performs validation.
```
dconnolly, line 246: 👍

frost-core/src/frost/keys.rs
```diff
@@ -327,7 +318,7 @@ pub fn keygen_with_dealer<C: Ciphersuite, R: RngCore + CryptoRng>(
     num_signers: u8,
     threshold: u8,
     mut rng: R,
-) -> Result<(Vec<SharePackage<C>>, PublicKeyPackage<C>), &'static str> {
+) -> Result<(Vec<SecretShare<C>>, PublicKeyPackage<C>), &'static str> {
```
dconnolly, line 321: 👍

frost-core/src/frost/keys.rs
```diff
@@ -364,7 +348,7 @@ pub fn keygen_with_dealer<C: Ciphersuite, R: RngCore + CryptoRng>(
 /// A FROST keypair, which can be generated either by a trusted dealer or using
 /// a DKG.
 ///
-/// When using a central dealer, [`SharePackage`]s are distributed to
+/// When using a central dealer, [`SecretShare`]s are distributed to
```
dconnolly, line 351: 👍

frost-core/src/frost/keys.rs
conradoplg: DKG can't use the same struct, because it requires more data to verify the share :cry: (at least I couldn't figure it out)

dconnolly: Gotcha

dconnolly (approved):
☺️

Issue #121, opened 2022-08-25
Organize the Re-randomizable FROST implementation


conradoplg:
Currently, it is composed of:

- A branch in `frost-core`
- A branch in `reddsa` that points to it

Some possible approaches for the `frost-core` part:

1. Incorporate the code in `frost-core` behind a feature flag. Upside: super simple; downside: we did not want to have re-randomizable stuff inside the crate
2. Create a new crate with the same code. Upside: no re-randomizable code in the original crate; downside: a LOT of duplicated code
3. Expose the  `frost-core` innards, maybe behind a feature flag, so that a new crate can be created without duplicating code. Upside: no re-randomizable code in the original crate, less duplicated code. Downside: exposed innards

Some possible approaches for the `reddsa` part:

1. Merge the generic implementation into `reddsa`, leave it for clients to parametrize it
2. The above but also add parametrized implementations to the  `redjubjub` and `redpallas` crate

Come up with a design and organize the existing code following it.

PR #83, opened 2022-07-11
Expose internals to support randomization

conradoplg:
Part of #121 along with https://github.com/ZcashFoundation/reddsa/pull/26

This implements the third option listed in #121. Some internal of frost-core are exposed under the "internals" feature (naming suggestions welcome) which allow implementing randomized FROST while reducing (but not eliminating) the amount of duplicated code.

PR #152, opened 2022-10-20
Expose internals to support randomization (simplified)

conradoplg:
Part of https://github.com/ZcashFoundation/frost/issues/121 along with https://github.com/ZcashFoundation/reddsa/pull/33

This is alternative to https://github.com/ZcashFoundation/frost/pull/83, which requires less change but is conceptually less "pure". The main thing that is required for randomized FROST is having the randomized group public key hashed when computing the binding factors. This introduces an internal function with a "additional_prefix" parameter that is empty in regular FROST and can be used to randomized FROST.

This allows less duplicated code in randomized FROST (basically just the high-level `sign` and `aggregate` functions).

While targeting `main`, this builds on top of https://github.com/ZcashFoundation/frost/pull/83, so you can see the last commit to check the differences between those.


frost-core/src/frost.rs
```diff
@@ -255,43 +279,41 @@ pub struct GroupCommitment<C: Ciphersuite>(pub(super) Element<C>);
...
 
-impl<C> TryFrom<&SigningPackage<C>> for GroupCommitment<C>
+/// Generates the group commitment which is published as part of the joint
+/// Schnorr signature.
+///
+/// Implements [`compute_group_commitment`] from the spec.
+///
+/// [`compute_group_commitment`]: https://www.ietf.org/archive/id/draft-irtf-cfrg-frost-10.html#section-4.5
+#[cfg_attr(feature = "internals", visibility::make(pub))]
+fn compute_group_commitment<C>(
```
dconnolly, line 289: Can we also call this function from the `TryFrom` impl for the default (`internals` feature not enabled) case?

frost-core/src/frost/round2.rs
```diff
@@ -109,6 +109,25 @@ where
...
+    challenge: Challenge<C>,
+) -> SignatureShare<C> {
+    let z_share: <<C::Group as Group>::Field as Field>::Scalar = signer_nonces.hiding.0
+        + (signer_nonces.binding.0 * binding_factor.0)
+        + (lambda_i * key_package.secret_share.0 * challenge.0);
+    let signature_share = SignatureShare::<C> {
+        identifier: *key_package.identifier(),
+        signature: SignatureResponse::<C> { z_share },
+    };
+    signature_share
```
dconnolly, line 128:
```suggestion
    let z_share: <<C::Group as Group>::Field as Field>::Scalar = signer_nonces.hiding.0
        + (signer_nonces.binding.0 * binding_factor.0)
        + (lambda_i * key_package.secret_share.0 * challenge.0);
        
    return SignatureShare::<C> {
        identifier: *key_package.identifier(),
        signature: SignatureResponse::<C> { z_share },
    }
```

frost-core/src/signature.rs
```diff
@@ -20,6 +20,15 @@ where
...
     <C::Group as Group>::Field: Field,
 {
+    /// Create a new Signature.
+    #[cfg(feature = "internals")]
+    pub fn new(
+        R: <C::Group as Group>::Element,
+        z: <<C::Group as Group>::Field as Field>::Scalar,
+    ) -> Self {
+        Self { R, z }
+    }
```
dconnolly, line 30: 👍

frost-core/src/verifying_key.rs
```diff
@@ -23,6 +23,18 @@ where
...
+    #[cfg(feature = "internals")]
+    pub fn new(element: <C::Group as Group>::Element) -> Self {
+        Self { element }
+    }
+
+    /// Return the underlying element.
+    #[cfg(feature = "internals")]
+    pub fn to_element(self) -> <C::Group as Group>::Element {
+        self.element
+    }
```
dconnolly, line 36: 👍

dconnolly (dismissed):
looking good, some comments

frost-core/src/frost.rs
conradoplg: We can, but it wouldn't be used. `compute_group_commitment` receives a previously-computed `binding_factor_list` (which _is_ previously computed everywhere it's called) so there is no motivation to actually call the TryFrom (which would recompute `binding_factor_list`). We could make a `TryFrom` from a tuple, but I think that's usually non-intuitive...

frost-core/src/frost/round2.rs
conradoplg: Done in 1b9efcf without the `return`

frost-rerandomized/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+/// Implements [`sign`] from the spec.
+///
+/// Receives the message to be signed and a set of signing commitments and a set
+/// of randomizing commitments to be used in that signing operation, including
+/// that for this participant.
+///
+/// Assumes the participant has already determined which nonce corresponds with
+/// the commitment that was assigned by the coordinator in the SigningPackage.
+///
+/// [`sign`]: https://www.ietf.org/archive/id/draft-irtf-cfrg-frost-10.html#name-round-two-signature-share-g
```
dconnolly, line 28:
```suggestion
/// [`sign`]: https://www.ietf.org/archive/id/draft-irtf-cfrg-frost-11.html#name-round-two-signature-share-g
```

frost-core/src/frost.rs
```diff
@@ -223,13 +232,18 @@ where
 
     /// Compute the preimages to H3 to compute the per-signer binding factors
```
dconnolly, line 233: s/H3/H1/ ?

https://github.com/ZcashFoundation/frost/blob/a02649d75685c20d972d15ff47ff04084d03fd8a/frost-core/src/frost.rs#L123

```suggestion
    /// Compute the preimages to H1 to compute the per-signer binding factors
```

frost-rerandomized/src/lib.rs
```diff
@@ -0,0 +1,209 @@
...
+        &self.randomizer_point
+    }
+
+    /// Return the randomized group public key.
+    ///
+    /// It can be used to verify the final signature.
+    pub fn randomized_group_public_key(&self) -> &frost_core::VerifyingKey<C> {
+        &self.randomized_group_public_key
+    }
+}
```
dconnolly, line 209: 👍

dconnolly (dismissed):
lgtm! very nice

dconnolly (approved):
lgtm! very nice

PR #154, opened 2022-10-21
Fix batching for P-256

conradoplg:
Depends on #114 

Uses the `little_endian_encoding` introduced in #114 to fix the bug described in #141.
Additionally, fixs another bug that would cause batching to fail for P-256, which was caused by the NAF size being fixed to 256. Since it's at most one element larger than the bit length, it worked for Ristretto (255) but not for P-256. I just increased it to 257.

Closes https://github.com/ZcashFoundation/frost/issues/141


dconnolly:
@mergify rebase

dconnolly (approved):
🎉

PR #157, opened 2022-10-21
use max_signers and min_signers instead of num_signers and threshold to better follow spec

conradoplg:
Depends on #156 because it touches the same code.

The current naming can be confusing because in the spec, num_participants is the number of participants creating a signature, while in the code it's the maximum number of participants.

This renames them to become more similar to the spec. However, I decided to keep "signer" instead of "participant" because we use that terminology in a bazillion other places. We can make the change in a separate PR if we decide to.

This is a cleanup without a matching issue.


dconnolly (approved):
max and min, does what it says on the tin

PR #159, opened 2022-10-21
Remove `as` keyword when not needed, use `Scalar`/`Element` shortcuts

conradoplg:
@upbqdn pointed out that some usages of the `as` keyword are not needed.

Additionally we weren't using the `Scalar` / `Element` type "shortcuts" everywhere applicable.

Cleanup without matching issue.

Depends on #114


dconnolly (approved):

frost-core/src/frost/keys.rs
```diff
@@ -364,14 +363,12 @@ fn evaluate_polynomial<C: Ciphersuite>(
 fn evaluate_vss<C: Ciphersuite>(
     commitment: &VerifiableSecretSharingCommitment<C>,
     identifier: Identifier<C>,
-) -> Result<<<C as Ciphersuite>::Group as Group>::Element, &'static str> {
+) -> Result<Element<C>, &'static str> {
+
```
dconnolly, line 367:
```suggestion
```

dconnolly (approved):

Issue #131, opened 2022-09-21
Improve examples in curve crates


conradoplg:
The curve-specific crates (frost-ristretto255, frost-p256) have examples that don't actually show how to use FROST.

Add a better example to them.

Consider changing `gendoc.py` from #130 to automatically generate the P-256 example from the Ristretto255 example, if needed.

## Design

It's difficult to exemplify a distributed protocol in a single-threaded example. I thought of using threads in the example, using channels to simulate communication. I think that could make it clearer for users how to use the library, but I'm not sure.

The alternative is to add a bunch of comments detailing where communication would take place, like in the test for #129

PR #160, opened 2022-10-24
Improve examples

conradoplg:
- Add a ciphersuite-specific example to each crate
- Add code to `gendoc` to also copy the example

Closes https://github.com/ZcashFoundation/frost/issues/131, but we might want to use async to make the example clearer and to also add a DKG example


frost-ristretto255/README.md
```diff
@@ -3,30 +3,98 @@ of signers (FROST).
...
 
 ```rust
+use std::{collections::HashMap, convert::TryFrom};
 use rand::thread_rng;
-use frost_ristretto255::*;
+use frost_ristretto255 as frost;
+
+let mut rng = thread_rng();
+let numsigners = 5;
+let threshold = 3;
```
conradoplg, line 20: I need to change this to min/max_signers and then also copy the example to #164 after that merges so I'll wait for that to happen and update this PR

conradoplg: Done

dconnolly (approved):
lgtm!

Should we run `gendoc` as part of CI? That way changes to `frost-ristretto255` rustdoc get applied to all others automatically

dconnolly:
> lgtm!
> 
> Should we run `gendoc` as part of CI? That way changes to `frost-ristretto255` rustdoc get applied to all others automatically

I wonder if we can make the job make suggested changes by the suggested change mechanic vs a new commit 🤔

conradoplg:
> > lgtm!
> > Should we run `gendoc` as part of CI? That way changes to `frost-ristretto255` rustdoc get applied to all others automatically
> 
> I wonder if we can make the job make suggested changes by the suggested change mechanic vs a new commit thinking

Yeah I'm not very fond of CI making commits. I think the easiest thing to do is making gendoc behave like rustfmt and return nonzero if it does makes changes, so that CI will fail if we forget to run it

dconnolly:
> > > lgtm!
> > > Should we run `gendoc` as part of CI? That way changes to `frost-ristretto255` rustdoc get applied to all others automatically
> > 
> > 
> > I wonder if we can make the job make suggested changes by the suggested change mechanic vs a new commit thinking
> 
> Yeah I'm not very fond of CI making commits. I think the easiest thing to do is making gendoc behave like rustfmt and return nonzero if it does makes changes, so that CI will fail if we forget to run it

I like this 👍

PR #172, opened 2022-10-27
use Error everywhere and add enums as needed

### LINKED ISSUE #167: Make Error strings proper Error types/enum variants
We already use `thiserror` and that may be enough.

Goal is to have proper error types instead of just strings all over the place

conradoplg:
Closes #167 

I decided to keep everything into a single enum to see how many items would it have.

It's a bit big but I'm not sure if it warrants having different enums for different parts. What do you think?

Also I put the `FromHex` impls behind the test flag since it's only used in tests, and did not create a specific error for them. Also not sure what is the best approach here.

### DIFF
--- a/frost-core/src/error.rs
+++ b/frost-core/src/error.rs
@@ -6,6 +6,15 @@ use thiserror::Error;
 #[non_exhaustive]
 #[derive(Error, Debug, Copy, Clone, Eq, PartialEq)]
 pub enum Error {
+    /// min_signers it invalid
+    #[error("min_signers must be at least 2 and not larger than max_signers")]
+    InvalidMinSigners,
+    /// max_signers it invalid
+    #[error("max_signers must be at least 2")]
+    InvalidMaxSigners,
+    /// max_signers it invalid
+    #[error("coefficients must have min_signers-1 elements")]
+    InvalidCoefficients,
     /// This identifier is unserializable.
     #[error("Malformed identifier is unserializable.")]
     MalformedIdentifier,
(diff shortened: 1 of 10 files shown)


dconnolly:
> It's a bit big but I'm not sure if it warrants having different enums for different parts. What do you think?

It's a bit big but I think it's fine, it's nice to just have `crate::Error` or `frost::Error` for everything and not a bunch of `Error` enums per module that we have to convert between or describe relationships between.

frost-core/src/error.rs
```diff
@@ -6,6 +6,15 @@ use thiserror::Error;
...
 #[derive(Error, Debug, Copy, Clone, Eq, PartialEq)]
 pub enum Error {
+    /// min_signers it invalid
+    #[error("min_signers must be at least 2 and not larger than max_signers")]
+    InvalidMinSigners,
+    /// max_signers it invalid
+    #[error("max_signers must be at least 2")]
+    InvalidMaxSigners,
+    /// max_signers it invalid
+    #[error("coefficients must have min_signers-1 elements")]
```
dconnolly, line 16:
```suggestion
    /// min_signers is invalid
    #[error("min_signers must be at least 2 and not larger than max_signers")]
    InvalidMinSigners,
    /// max_signers is invalid
    #[error("max_signers must be at least 2")]
    InvalidMaxSigners,
    /// max_signers is invalid
    #[error("coefficients must have min_signers-1 elements")]
```

frost-core/src/frost.rs
```diff
@@ -117,6 +117,7 @@ where
     }
 }
 
+#[cfg(any(test, feature = "test-impl"))]
```
dconnolly, line 120: 👍

frost-core/src/frost/keys/dkg.rs
```diff
@@ -235,17 +233,17 @@ fn compute_verifying_keys<C: Ciphersuite>(
...
+                        .ok_or(Error::PackageNotFound)?
                         .commitment,
                 )
             })
             // Chain our own commitment vector
             .chain(iter::once(Ok(&round2_secret_package.commitment)))
         {
-            let result = evaluate_vss(commitments?, i)?;
+            let result = evaluate_vss(commitments?, i);
             y_i = y_i + result;
```
dconnolly, line 247:
```suggestion
            y_i = y_i + evaluate_vss(commitments?, i);
```

dconnolly (commented):
Looking good, one or two small suggestions

dconnolly (approved):
🎉

PR #176, opened 2022-11-02
remove random_nonzero from Field trait

conradoplg:
This is a optional cleanup.

Upside: removes a trait method
Downside: implementers must guarantee that PartialEq/Eq for scalars are constant time.

If that does not look like a good idea, we can change this PR to simply call the global `random_nonzero` function in each ciphersuite just to avoid duplicated code, making it possible for other ciphersuites to provide their own implementation if needed.

### DIFF
--- a/frost-core/src/frost/keys.rs
+++ b/frost-core/src/frost/keys.rs
@@ -12,7 +12,10 @@ use hex::FromHex;
 use rand_core::{CryptoRng, RngCore};
 use zeroize::{DefaultIsZeroes, Zeroize};
 
-use crate::{frost::Identifier, Ciphersuite, Element, Error, Field, Group, Scalar, VerifyingKey};
+use crate::{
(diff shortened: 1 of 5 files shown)


dconnolly (approved):
I like this; we can explore doing nice constant-time impls of `Eq/PartialEq` in our top level crates, and whether we can do this by default in `-core` by impl'ing `ConstantTimeEq` and calling that from `Eq/PartialEq` for our generic `Scalar`s

conradoplg:
> I like this; we can explore doing nice constant-time impls of `Eq/PartialEq` in our top level crates, and whether we can do this by default in `-core` by impl'ing `ConstantTimeEq` and calling that from `Eq/PartialEq` for our generic `Scalar`s

I just checked and we can get away with simply adding `ConstantTimeEq` as a bound in `Field::Scalar`. All our supported curves already implement it for their scalar fields.

PR #182, opened 2022-11-22
add signer identifier to InvalidSignatureShare as a byte array

conradoplg:
One feature of FROST is that cheaters can be detected which allows removing/blocking them if desired.

However, currently, there is now way to identify which participant caused the signature aggregation to fail.

This adds the signer identifier to the InvalidSignatureShare error, which allows this.

Ideally we'd add a `Identifier` instead of a hex string. But since identifier is generic on the ciphersuite, that would require making the Error enum also generic, but that doesn't work on the Field trait (which has two methods that return Error) since it isn't aware of the ciphersuite it's in.

We could e.g. split the error enum, but this is the simplest approach. I went ahead and created the PR since it makes a bit clearer to think about the problem, but I'd be happy to use another approach if preferred. **Edit: I did the alternate approach in https://github.com/ZcashFoundation/frost/pull/183**

There is not matching issue for this, I can create it if required.

### DIFF
--- a/frost-core/src/error.rs
+++ b/frost-core/src/error.rs
@@ -55,8 +55,9 @@ pub enum Error {
     #[error("Invalid signature share.")]
     InvalidSignatureShare {
         /// The identifier of the signer whose share validation failed,
-        /// encoded as a little-endian byte string in hex format.
-        signer: String,
+        /// encoded as a byte vector with ciphersuite-dependent endianness
+        /// (can be decoded with [`Identifier::deserialize`]).
+        signer: Vec<u8>,
     },
     /// Secret share verification failed.
     #[error("Invalid secret share.")]
--- a/frost-core/src/error.rs
+++ b/frost-core/src/error.rs
@@ -4,7 +4,7 @@ use thiserror::Error;
 
 /// An error related to FROST.
 #[non_exhaustive]
-#[derive(Error, Debug, Copy, Clone, Eq, PartialEq)]
+#[derive(Error, Debug, Clone, Eq, PartialEq)]
 pub enum Error {
     /// min_signers is invalid
     #[error("min_signers must be at least 2 and not larger than max_signers")]
@@ -53,7 +53,11 @@ pub enum Error {
     IdentityCommitment,
     /// Signature share verification failed.
     #[error("Invalid signature share.")]
-    InvalidSignatureShare,
+    InvalidSignatureShare {
+        /// The identifier of the signer whose share validation failed,
+        /// encoded as a little-endian byte string in hex format.
+        signer: String,
+    },
     /// Secret share verification failed.
     #[error("Invalid secret share.")]
     InvalidSecretShare,
--- a/frost-core/src/error.rs
+++ b/frost-core/src/error.rs
@@ -56,7 +56,7 @@ pub enum Error {
     InvalidSignatureShare {
         /// The identifier of the signer whose share validation failed,
         /// encoded as a byte vector with ciphersuite-dependent endianness
-        /// (can be decoded with [`Identifier::deserialize`]).
+        /// (can be decoded with [`crate::frost::Identifier::deserialize`]).
         signer: Vec<u8>,
     },
     /// Secret share verification failed.
--- a/frost-core/src/frost/identifier.rs
+++ b/frost-core/src/frost/identifier.rs
@@ -19,9 +19,22 @@ impl<C> Identifier<C>
 where
     C: Ciphersuite,
 {
-    // Serialize the underlying scalar.
-    pub(crate) fn serialize(&self) -> <<C::Group as Group>::Field as Field>::Serialization {
+    /// Serialize the identifier using the ciphersuite encoding.
+    pub fn serialize(&self) -> <<C::Group as Group>::Field as Field>::Serialization {
         <<C::Group as Group>::Field>::serialize(&self.0)
+    }
+
+    /// Deserialize an Identifier from a serialized buffer.
+    /// Returns an error if it attempts to deserialize zero.
+    pub fn deserialize(
+        buf: &<<C::Group as Group>::Field as Field>::Serialization,
+    ) -> Result<Self, Error> {
+        let scalar = <<C::Group as Group>::Field>::deserialize(buf)?;
+        if scalar == <<C::Group as Group>::Field>::zero() {
+            Err(Error::InvalidZeroScalar)
+        } else {
+            Ok(Self(scalar))
+        }
     }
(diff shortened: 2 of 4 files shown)


frost-core/src/frost/identifier.rs
```diff
@@ -25,6 +25,23 @@ where
...
+{
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(
+            f,
+            "{}",
+            hex::encode(
+                <<<C as Ciphersuite>::Group as Group>::Field as Field>::little_endian_serialize(
+                    &self.0
+                )
+            )
```
conradoplg, line 40: I'm a bit keen into using big endian instead because it reads more naturally as a string, but I don't want Yet Another Endianness Discussion, so I'm happy with whatever :rofl:

dconnolly: Little endian! Littlest endians!

dconnolly: (
image
)

frost-core/src/error.rs
```diff
@@ -53,7 +53,11 @@ pub enum Error {
     IdentityCommitment,
     /// Signature share verification failed.
     #[error("Invalid signature share.")]
-    InvalidSignatureShare,
+    InvalidSignatureShare {
+        /// The identifier of the signer whose share validation failed,
+        /// encoded as a little-endian byte string in hex format.
+        signer: String,
```
dconnolly, line 59: If this is little-endian bytes, should it be just Vec<u8> rather than String? If String makes things easier that's a fine answer

frost-core/src/tests.rs
```diff
@@ -59,6 +59,36 @@ pub fn check_sign_with_dealer<C: Ciphersuite, R: RngCore + CryptoRng>(mut rng: R
...
+fn check_corrupted_share<C: Ciphersuite + PartialEq, R: RngCore + CryptoRng>(
+    signature_shares: &[frost::round2::SignatureShare<C>],
+    mut rng: R,
+    signing_package: &frost::SigningPackage<C>,
+    pubkeys: &frost::keys::PublicKeyPackage<C>,
+) {
+    let identifier_one = 1u16.try_into().expect("should be nonzero");
+    let mut corrupted_signature_shares = signature_shares.to_owned();
+    let random_share = <<C::Group as Group>::Field>::random(&mut rng);
+    corrupted_signature_shares
```
dconnolly, line 73:
```suggestion

    corrupted_signature_shares
```

frost-core/src/tests.rs
```diff
@@ -59,6 +59,36 @@ pub fn check_sign_with_dealer<C: Ciphersuite, R: RngCore + CryptoRng>(mut rng: R
...
+        .iter_mut()
+        .find(|share| share.identifier == identifier_one)
+        .unwrap()
+        .signature
+        .z_share = random_share;
+    let group_signature_res =
```
dconnolly, line 79:
```suggestion

    let group_signature_res =
```

frost-core/src/tests.rs
```diff
@@ -59,6 +59,36 @@ pub fn check_sign_with_dealer<C: Ciphersuite, R: RngCore + CryptoRng>(mut rng: R
...
+        frost::aggregate(signing_package, &corrupted_signature_shares[..], pubkeys);
+    match group_signature_res {
```
dconnolly, line 81:
```suggestion

    match group_signature_res {
```

frost-core/src/tests.rs
```diff
@@ -59,6 +59,36 @@ pub fn check_sign_with_dealer<C: Ciphersuite, R: RngCore + CryptoRng>(mut rng: R
...
+        Ok(_) => panic!("should fail"),
+        Err(Error::InvalidSignatureShare { signer }) => {
+            // starts_with is used instead of equality so that this works even with larger scalar fields
+            assert!(signer
+                .starts_with("0100000000000000000000000000000000000000000000000000000000000000"));
+        }
```
dconnolly, line 87: Interesting; should we have a `Signer` type that impls a method that kinda wraps this up nicer?

frost-core/src/error.rs
conradoplg: Good point, no particular reason to use a string.

I'm kinda having second thoughts about the endianess issue. Should we use the ciphersuite-specified endianess (i.e. just call `serialize`) so that the caller can deserialize it if they wish?

conradoplg: OK I had a shot at this in 530b5ee

- Switches to `Vec<u8>`
- Uses ciphersuite-endianess
- Introduces `Identifier::deserialize`
- Makes `Identifier::serialize` public (my rationale is that people will need to serialize/deserialize identifiers for communicating)

Keeping it little-endian would require adding `Identifier::(de)serialize_little_endian` which is more awkward.

What do you think?

dconnolly:
> 

I like the approach in #183, it works nicely

frost-core/src/tests.rs
dconnolly: Ah yeah a new `Signer` type would overlap a lot with `Identifier`; what I was thinking was either way have a type that we could impl a comparison (either with bytes, or this sort of encoding as a string) explicitly (maybe impl `Eq` etc) vs having the signer be a `String` and calling `starts_with`. 

If we go with #183 we can just do a more fleshed out `impl Eq for Identifier` if we like, or a test-only / internal `impl Identifier<C> { pub fn starts_with(&self) -> bool { ... } }`, or something

dconnolly:
So if we go with #183, should this get refactored on top of that?

frost-core/src/error.rs
```diff
@@ -53,7 +53,12 @@ pub enum Error {
     IdentityCommitment,
     /// Signature share verification failed.
     #[error("Invalid signature share.")]
-    InvalidSignatureShare,
+    InvalidSignatureShare {
+        /// The identifier of the signer whose share validation failed,
+        /// encoded as a byte vector with ciphersuite-dependent endianness
+        /// (can be decoded with [`crate::frost::Identifier::deserialize`]).
```
dconnolly, line 59: 👍

conradoplg (commented):
> I like the approach in #183, it works nicely

Cool!

> So if we go with #183, should this get refactored on top of that?

This is an alternative implementation, so we can disregard this PR, but the Identifier (de)serialization that I ended up adding are useful so I'll add those to #183

conradoplg:
Closing in favor of https://github.com/ZcashFoundation/frost/pull/183

PR #183, opened 2022-11-22
add Identifier to InvalidSignatureShare and split Error type

conradoplg:
This is an alternative to #182 which is cleaner but more invasive. I had to split the Error enum into Error, FieldError and GroupError.

I added the `Debug` bound to the ciphersuite because assertions need it. Alternatively we can use `debugless_unwrap()` which we already use elsewhere, and remove the bound.


frost-core/src/error.rs
```diff
@@ -53,7 +40,10 @@ pub enum Error {
     IdentityCommitment,
     /// Signature share verification failed.
     #[error("Invalid signature share.")]
-    InvalidSignatureShare,
+    InvalidSignatureShare {
+        /// The identifier of the signer whose share validation failed.
+        signer: Identifier<C>,
+    },
```
dconnolly, line 46: 👍

frost-core/src/error.rs
```diff
@@ -72,4 +62,37 @@ pub enum Error {
...
+#[non_exhaustive]
+#[derive(Error, Debug, Copy, Clone, Eq, PartialEq)]
+pub enum FieldError {
+    /// The encoding of a group scalar was malformed.
+    #[error("Malformed scalar encoding.")]
+    MalformedScalar,
+    /// This scalar MUST NOT be zero.
+    #[error("Invalid for this scalar to be zero.")]
+    InvalidZeroScalar,
+}
```
dconnolly, line 83: 👍

frost-core/src/error.rs
```diff
@@ -72,4 +62,37 @@ pub enum Error {
...
+
+/// An error related to an elliptic curve Group.
```
dconnolly, line 85:
```suggestion
/// An error related to a Group or one of its Elements.
```

frost-core/src/frost.rs
```diff
@@ -44,8 +44,10 @@ where
     /// Deserializes [`BindingFactor`] from bytes.
     pub fn from_bytes(
         bytes: <<C::Group as Group>::Field as Field>::Serialization,
-    ) -> Result<Self, Error> {
-        <<C::Group as Group>::Field>::deserialize(&bytes).map(|scalar| Self(scalar))
+    ) -> Result<Self, Error<C>> {
+        <<C::Group as Group>::Field>::deserialize(&bytes)
+            .map(|scalar| Self(scalar))
+            .map_err(|e| e.into())
```
dconnolly, line 50: Is this needed because the original `e` may be `FieldError` or `GroupError`?

frost-core/src/frost/identifier.rs
```diff
@@ -109,11 +109,11 @@ impl<C> TryFrom<u16> for Identifier<C>
 where
     C: Ciphersuite,
 {
-    type Error = Error;
+    type Error = Error<C>;
 
     fn try_from(n: u16) -> Result<Identifier<C>, Self::Error> {
         if n == 0 {
-            Err(Self::Error::InvalidZeroScalar)
+            Err(FieldError::InvalidZeroScalar.into())
```
dconnolly, line 116: 👍

frost-core/src/frost/keys.rs
```diff
@@ -1,4 +1,5 @@
 //! FROST keys, keygen, key shares
+#![allow(clippy::type_complexity)]
```
dconnolly, line 2: 😅

frost-core/src/frost/round2.rs
```diff
@@ -86,11 +88,13 @@ where
...
         challenge: &Challenge<C>,
-    ) -> Result<(), Error> {
+    ) -> Result<(), Error<C>> {
         if (<C::Group>::generator() * self.signature.z_share)
             != (group_commitment_share.0 + (public_key.0 * challenge.0 * lambda_i))
         {
-            return Err(Error::InvalidSignatureShare);
+            return Err(Error::InvalidSignatureShare {
+                signer: self.identifier,
+            });
```
dconnolly, line 97: 🎉

dconnolly (approved):
Looks great! 😎

dconnolly (changes requested):
Looks great, but may have missed some `C`s 😅

image

conradoplg:
> Looks great, but may have missed some `C`s sweat_smile
> 
> image

Oops, that's because `secp256k1` was merged after I created this PR. I've just updated it.

frost-core/src/frost.rs
conradoplg: Exactly, in this case it's always `FieldError`, but we want an `Error`. Since `Error` has a `FieldError(#[from] FieldError),` it has a `From` implementation that allows us to just call `.into()`.

conradoplg:
@dconnolly I added the Identifier serialization methods here too, and updated it to work with the recently merged Ed448. This should be good to go, and we can close #182

frost-core/src/frost/identifier.rs
```diff
@@ -19,10 +19,23 @@ impl<C> Identifier<C>
...
+    pub fn serialize(&self) -> <<C::Group as Group>::Field as Field>::Serialization {
         <<C::Group as Group>::Field>::serialize(&self.0)
     }
+
+    /// Deserialize an Identifier from a serialized buffer.
+    /// Returns an error if it attempts to deserialize zero.
+    pub fn deserialize(
+        buf: &<<C::Group as Group>::Field as Field>::Serialization,
+    ) -> Result<Self, Error<C>> {
+        let scalar = <<C::Group as Group>::Field>::deserialize(buf)?;
```
dconnolly, line 32:
```suggestion

        let scalar = <<C::Group as Group>::Field>::deserialize(buf)?;
```

dconnolly: lol, this is against rustfmt

image

frost-core/src/frost/identifier.rs
```diff
@@ -19,10 +19,24 @@ impl<C> Identifier<C>
...
     }
+
+    /// Deserialize an Identifier from a serialized buffer.
+    /// Returns an error if it attempts to deserialize zero.
+    pub fn deserialize(
+        buf: &<<C::Group as Group>::Field as Field>::Serialization,
+    ) -> Result<Self, Error<C>> {
+
+        let scalar = <<C::Group as Group>::Field>::deserialize(buf)?;
+        if scalar == <<C::Group as Group>::Field>::zero() {
```
dconnolly, line 34:
```suggestion
        let scalar = <<C::Group as Group>::Field>::deserialize(buf)?;
        
        if scalar == <<C::Group as Group>::Field>::zero() {
```

frost-core/src/frost/identifier.rs
```diff
@@ -19,10 +19,24 @@ impl<C> Identifier<C>
...
     }
+
+    /// Deserialize an Identifier from a serialized buffer.
+    /// Returns an error if it attempts to deserialize zero.
+    pub fn deserialize(
+        buf: &<<C::Group as Group>::Field as Field>::Serialization,
+    ) -> Result<Self, Error<C>> {
+        let scalar = <<C::Group as Group>::Field>::deserialize(buf)?;
+        
+        if scalar == <<C::Group as Group>::Field>::zero() {
```
dconnolly, line 34:
```suggestion
        let scalar = <<C::Group as Group>::Field>::deserialize(buf)?;
        if scalar == <<C::Group as Group>::Field>::zero() {
```

dconnolly (approved):

PR #184, opened 2022-11-22
add gendoc --check, use in CI

### LINKED ISSUE #174: Run `gendoc` as a pass/fail as part of CI
Alá `rustfmt` or `clippy`, gives feedback without making edits.

conradoplg:
Closes #174 

This also fixes gendoc to work with secp256k1 (another reason why we need this PR :laughing: )

This is failing on purpose to show it works. After it's reviewed I'll commit the updated files

### DIFF
--- a/frost-secp256k1/src/lib.rs
+++ b/frost-secp256k1/src/lib.rs
@@ -242,10 +242,10 @@ impl Ciphersuite for Secp256K1Sha256 {
     }
 }
 
-type R = Secp256K1Sha256;
+type S = Secp256K1Sha256;
 
 /// A FROST(secp256k1, SHA-256) participant identifier.
-pub type Identifier = frost::Identifier<R>;
+pub type Identifier = frost::Identifier<S>;
 
 /// FROST(secp256k1, SHA-256) keys, key generation, key shares.
 pub mod keys {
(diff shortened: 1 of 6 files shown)


dconnolly:
Failed?

frost-secp256k1/src/lib.rs
```diff
@@ -268,21 +268,101 @@ pub mod keys {
...
+        use super::*;
+
+        /// The secret package that must be kept in memory by the participant
+        /// between the first and second parts of the DKG protocol (round 1).
+        ///
+        /// # Security
+        ///
+        /// This package MUST NOT be sent to other participants!
+        pub type Round1SecretPackage = frost::keys::dkg::Round1SecretPackage<S>;
+
```
dconnolly, line 298: At this point, should this live in its own file (`dkg.rs`) ?

conradoplg:
@dconnolly I ran gendoc and committed the result, and it's now passing as expected

frost-secp256k1/src/lib.rs
dconnolly: Tracked in: https://github.com/ZcashFoundation/frost/issues/185

dconnolly (approved):
🎉

PR #187, opened 2022-12-01
add Ed448 support

### LINKED ISSUE #71: Add Ed448 support
We want to support all curves defined in the spec.

conradoplg:
Closes #71 

This is currently pointing `Ed448-Goldilocks` to the branch in our fork corresponding to the PR https://github.com/crate-crypto/Ed448-Goldilocks/pull/23 . It will need to be updated it upstream merges or if we decided to merge in our fork.

This also removes the `Default` bound from `Serialization` because it's not strictly needed and makes it harder to support Ed448, since arrays larger than 256 bytes do not implement it.


frost-ed448/tests/frost.rs
```diff
@@ -0,0 +1,44 @@
...
+
+#[test]
+fn check_sign_with_dkg() {
+    let rng = thread_rng();
+
+    frost_core::tests::check_sign_with_dkg::<Ed448Shake256, _>(rng);
+}
+
+// TODO: make batching work for larger scalars
+// #[test]
```
dconnolly, line 21: 👍

frost-ed448/tests/frost.rs
```diff
@@ -0,0 +1,44 @@
...
+#[allow(unused)]
+fn check_batch_verify() {
+    let rng = thread_rng();
+
+    frost_core::tests::batch::batch_verify::<Ed448Shake256, _>(rng);
+}
+
+// TODO: make batching work for larger scalars
+// #[test]
```
dconnolly, line 30: 👍

frost-core/src/signature.rs
```diff
@@ -34,8 +34,9 @@ where
 
         let R_serialization = &R_bytes.try_into().map_err(|_| Error::MalformedSignature)?;
 
+        let one = <<C::Group as Group>::Field as Field>::one();
```
dconnolly, line 37: Why `one`, and not `zero()`? That's basically what `Serialization::default()` did before

frost-ed448/src/lib.rs
```diff
@@ -0,0 +1,405 @@
...
+        }
+    }
+
+    fn random<R: RngCore + CryptoRng>(rng: &mut R) -> Self::Scalar {
+        Scalar::random(rng)
+    }
+
+    fn serialize(scalar: &Self::Scalar) -> Self::Serialization {
+        let bytes = scalar.to_bytes();
+        std::array::from_fn(|i| if i < 56 { bytes[i] } else { 0 })
```
dconnolly, line 53: Is this a workaround until the upstream fixes are merged?

frost-ed448/src/lib.rs
```diff
@@ -0,0 +1,405 @@
...
+    fn identity() -> Self::Element {
+        Self::Element::identity()
+    }
+
+    fn generator() -> Self::Element {
+        Self::Element::generator()
+    }
+
+    fn serialize(element: &Self::Element) -> Self::Serialization {
+        element.compress().0
```
dconnolly, line 92: Funny they don't have a proper means of getting these bytes out

dconnolly (approved):
🎉

A comment or two but no blockers 💃

frost-ed448/src/lib.rs
conradoplg: Ed448 is tricky because points are encoded as 57 bytes (56 full bytes for the field element, forcing an extra byte for the sign). But scalars can be encoded as 56 bytes. I'm working on a PR upstream to make `Scalar::to_bytes()` return 57 bytes but it does make it weird because Ed448-Goldilocks also supports X448 and in that scenario there is no reason to encode scalars as 57 bytes and all test vectors have 56 bytes :smiling_face_with_tear: 

Anyway I'll check with the maintainer what seems to be the best solution, feedback is welcome

frost-ed448/src/lib.rs
conradoplg: To be fair `curve25519-dalek`, while it has `to/as_bytes`, also makes the bytes `pub`.

I want to eventually try to make the `Ed448-Goldilocks` API as similar as `curve25519-dalek` as possible

frost-core/src/signature.rs
conradoplg: Good point, no idea why I picked `one()`. Changed to  `zero()` in 17c80c1

frost-ed448/src/lib.rs
dconnolly: Weird, not my favorite, but oh well!

frost-core/src/signature.rs
dconnolly: Nice 😎

dconnolly (approved):
looks good! Thanks for answering my questions :D

PR #192, opened 2022-12-07
add bound to ConstantTimeEq in Scalar

conradoplg:
Not sure if we want this or not, but this allows us to use `ct_eq` where needed instead of just documenting that `eq` should be constant time.

(Ideally we could remove `Eq` to make sure we're using `ct_eq` where needed but this propagates everywhere and doesn't seem worth it)

### DIFF
--- a/frost-core/Cargo.toml
+++ b/frost-core/Cargo.toml
@@ -23,6 +23,7 @@ digest = "0.10"
 hex = { version = "0.4.3", features = ["serde"] }
 rand_core = "0.6"
 serde = { version = "1", optional = true, features = ["derive"] }
+subtle = "2.4.1"
 thiserror = "1.0"
 zeroize = { version = "1.5.4", default-features = false, features = ["derive"] }
 
--- a/frost-core/src/lib.rs
+++ b/frost-core/src/lib.rs
@@ -26,6 +26,7 @@ mod verifying_key;
 pub use error::Error;
 pub use signature::Signature;
 pub use signing_key::SigningKey;
+pub use subtle::ConstantTimeEq;
 pub use verifying_key::VerifyingKey;
 
 /// A prime order finite field GF(q) over which all scalar values for our prime order group can be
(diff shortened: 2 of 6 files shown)


dconnolly:
IIUC, this would require the `Scalar` to implement `subtle::ConstantTimeEq` directly, which some underlying curve/group libraries do, but not all. If the implementor of, say `frost-jq255s` imports `jq255s::Scalar`, because of the orphan rule, they can't impl `subtle::ConstantTimeEq` directly for `Scalar`, and would have to create a newtype wrapper just to fulfill this requirement

conradoplg:
> IIUC, this would require the `Scalar` to implement `subtle::ConstantTimeEq` directly, which some underlying curve/group libraries do, but not all. If the implementor of, say `frost-jq255s` imports `jq255s::Scalar` which does not impl this trait, because of the orphan rule, they can't impl `subtle::ConstantTimeEq` directly for `Scalar`, and would have to create a newtype wrapper just to fulfill this requirement

Yep that's the downside, one argument in favor is that all existing ciphersuites already implement it. But I don't feel strongly either way, feel free to close or accept it.

dconnolly:
> > IIUC, this would require the `Scalar` to implement `subtle::ConstantTimeEq` directly, which some underlying curve/group libraries do, but not all. If the implementor of, say `frost-jq255s` imports `jq255s::Scalar` which does not impl this trait, because of the orphan rule, they can't impl `subtle::ConstantTimeEq` directly for `Scalar`, and would have to create a newtype wrapper just to fulfill this requirement
> 
> Yep that's the downside, one argument in favor is that all existing ciphersuites already implement it. But I don't feel strongly either way, feel free to close or accept it.


True, I'm mindful of other projects who have expressed interest in building with `frost-core` who use different curves, so I don't want to encumber them.

Also secretly, a `Scalar` ideally would impl `Eq` as a call to `self.ct_eq()` under the hood anyway, so :P

PR #197, opened 2022-12-09
fix Identifier ordering; add big identifier test vectors

### LINKED ISSUE #93: Add tests for large indices
See #90

conradoplg:
Closes #93 

I was just adding the test for large identifiers and I found out that I introduced a bug :grimacing: 

Identifiers were being compared by serializing into little endian and comparing the arrays. But that's wrong; it needs to compare the big endian arrays.

This also fixes it.

I'll also create a PR on the Sage code that generates the vectors, but I'm thinking on doing that later so it doesn't get in the way of the spec work.

### DIFF
--- a/frost-core/src/frost/identifier.rs
+++ b/frost-core/src/frost/identifier.rs
@@ -57,7 +57,12 @@ where
     fn cmp(&self, other: &Self) -> std::cmp::Ordering {
         let serialized_self = <<C::Group as Group>::Field>::little_endian_serialize(&self.0);
         let serialized_other = <<C::Group as Group>::Field>::little_endian_serialize(&other.0);
-        serialized_self.as_ref().cmp(serialized_other.as_ref())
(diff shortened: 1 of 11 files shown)


frost-core/src/frost/identifier.rs
```diff
@@ -57,7 +57,12 @@ where
     fn cmp(&self, other: &Self) -> std::cmp::Ordering {
         let serialized_self = <<C::Group as Group>::Field>::little_endian_serialize(&self.0);
         let serialized_other = <<C::Group as Group>::Field>::little_endian_serialize(&other.0);
-        serialized_self.as_ref().cmp(serialized_other.as_ref())
+        // The default cmp uses lexicographic order; so we need the elements in big endian
```
dconnolly, line 60: Aha! lol

dconnolly (approved):
Nice!

PR #207, opened 2022-12-16
improve aggregate performance by verifying the aggregate signature first

conradoplg:
Implements https://github.com/cfrg/draft-irtf-cfrg-frost/pull/360

### DIFF
--- a/frost-core/src/frost.rs
+++ b/frost-core/src/frost.rs
@@ -323,11 +323,8 @@ where
 // Aggregation
 ////////////////////////////////////////////////////////////////////////////////
 
-/// Verifies each participant's signature share, and if all are valid,
-/// aggregates the shares into a signature to publish.
-///
-/// Resulting signature is compatible with verification of a plain SpendAuth
-/// signature.
+/// Aggregates the signature shares to produce a final signature that
+/// can be verified with the group public key.
 ///
 /// This operation is performed by a coordinator that can communicate with all
 /// the signing participants before publishing the final signature. The
--- a/frost-core/src/frost.rs
+++ b/frost-core/src/frost.rs
@@ -323,42 +323,8 @@ pub fn aggregate<C>(
 where
     C: Ciphersuite,
 {
-    // Encodes the signing commitment list produced in round one as part of generating [`BindingFactor`], the
-    // binding factor.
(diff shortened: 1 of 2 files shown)


frost-core/src/frost.rs
```diff
@@ -396,8 +366,53 @@ where
...
         z,
-    })
+    };
+
+    // Verify the aggregate signature
+    let verification_result = pubkeys
+        .group_public
+        .verify(signing_package.message(), &signature);
+
+    // If verification failed; verify each share to find the cheater.
```
upbqdn, line 379: The "only if" construct might be handy here:
```suggestion
    // Only if the verification of the aggregate signature failed; verify each share to find the cheater.
```

upbqdn (changes requested):
PR looks great. I noticed an outdated comment that I think we should change before merging this PR, so I'm requesting changes.

See https://github.com/ZcashFoundation/frost/blob/78e67d175b71deae26dec6ab608918e1e55b0234/frost-core/src/frost.rs#L326-L327

frost-core/src/frost.rs
```diff
@@ -396,8 +366,53 @@ where
...
+    // This approach is more efficient since we don't need to verificate all shares
```
dconnolly, line 380:
```suggestion
    // This approach is more efficient since we don't need to verify all shares
```

dconnolly (dismissed):
🎉

dconnolly (approved):

upbqdn (commented):
Looks great. However, I still think we should update this comment. https://github.com/ZcashFoundation/frost/blob/78e67d175b71deae26dec6ab608918e1e55b0234/frost-core/src/frost.rs#L326-L327

Unfortunately, GitHub won't let me make a suggestion.

conradoplg:
> Looks great. However, I still think we should update this comment.
> 
> https://github.com/ZcashFoundation/frost/blob/78e67d175b71deae26dec6ab608918e1e55b0234/frost-core/src/frost.rs#L326-L327
> 
> Unfortunately, GitHub won't let me make a suggestion.

Good catch, fixed it in a904af6, I also removed some outdated wording

upbqdn (approved):

PR #208, opened 2022-12-16
add ed25519 interoperability test

conradoplg:
After adding a similar test for [Jubjub and Pallas](https://github.com/ZcashFoundation/reddsa/pull/33/), I added a test to ensure that FROST-generate sigs can be validated by other crates.

I had great ambitions of adding this for every ciphersuite but then realized:

- Couldn't find Ed448 sig verification (Ed448-Goldilocks doesn't have it yet)
- Existing Ristretto Schnorr impls ([starsig](https://docs.rs/starsig/latest/starsig/), [schnorrkel](https://docs.rs/schnorrkel/latest/schnorrkel/index.html)) use fancy transcripts which are incompatible
- No existing P-256 Schnorr implementation (I guess we just invented it? :rofl: )
- No existing secp256k1 Schnorr implementations (Taproot is stricter)

I didn't look too hard though so we might revisit this.

### DIFF
--- a/frost-ed25519/tests/frost.rs
+++ b/frost-ed25519/tests/frost.rs
@@ -1,20 +1,50 @@
-use curve25519_dalek::{edwards::EdwardsPoint, traits::Identity};
 use frost_core::{Ciphersuite, Group, GroupError};
 use frost_ed25519::*;
+
+use curve25519_dalek::{edwards::EdwardsPoint, traits::Identity};
+use ed25519_dalek::Verifier;
 use rand::thread_rng;
+
+fn verify_signature(
+    msg: &[u8],
+    group_signature: frost_core::Signature<Ed25519Sha512>,
+    group_pubkey: frost_core::VerifyingKey<Ed25519Sha512>,
+) {
+    let sig = {
+        let bytes: [u8; 64] = group_signature.to_bytes();
+        ed25519_dalek::Signature::from(bytes)
+    };
+    let pub_key = {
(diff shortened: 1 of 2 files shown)


frost-ed25519/tests/frost.rs
```diff
@@ -1,20 +1,50 @@
...
+    // Check that signature validation has the expected result.
+    assert!(pub_key.verify(msg, &sig).is_ok());
+}
+
 #[test]
 fn check_sign_with_dealer() {
     let rng = thread_rng();
 
-    frost_core::tests::check_sign_with_dealer::<Ed25519Sha512, _>(rng);
+    // For the interoperability test it's better to test with multiple signatures
```
upbqdn, line 29: Why is it better to test multiple signatures with the dealer but not with the DKG?

upbqdn (approved):
Looks good. I left one question in the review.

frost-ed25519/tests/frost.rs
conradoplg: My reasoning was that key generation does not really impact this, only the signature generation itself. It's better to generate multiple signatures and verify them, but how the key was generated is not that important. I chose the trusted dealer test because it's faster.

Thinking about this though, I think it wouldn't hurt to test the DKG itself multiple times, so I added a loop in 37f867f

upbqdn: Makes sense, thanks for the clarification. I thought there was a deeper reason.

upbqdn (approved):

Issue #231, opened 2023-02-09
Improve accuracy of coverage reports


natalieesk:
Currently we are including helper functions, benchmark functions and gendoc functions in our code coverage report which skews the result.

Use the flag `--ignore-filename-regex` in the coverage.yaml file to exclude the relevant files.

mpguerra:
done in #232

PR #232, opened 2023-02-09
Improve accuracy of coverage reports 231

natalieesk:
#231


conradoplg (approved):
Looks good, thanks! 
I've been neglecting coverage results and this helps :+1:

conradoplg:
Ah BTW, since the title of the PR becomes the commit message (since we squash PRs), and we don't usually add the issue number to it (it's sufficient to link to it in the body of the PR), it's better to not include it just for consistency (not a big ideal though, and it has already been merged :laughing: )

PR #244, opened 2023-02-15
List of dependencies for audit

mpguerra:
Starting a list here as an example. If we don't think we want to put it in the same sort of place as Zebra, let's move it elsewhere


conradoplg:
I think that's pretty much it :+1: 

One thing I'm wondering about are the underlying crates with the elliptic curve implementations. In particular `ed448-goldilocks`  which is not as mature as the other ones. Is this something that we're willing to include in the audit?

mpguerra:
> One thing I'm wondering about are the underlying crates with the elliptic curve implementations. In particular `ed448-goldilocks` which is not as mature as the other ones. Is this something that we're willing to include in the audit?

Maybe we can do a partial audit of the code that is used by FROST

mpguerra:
Actually, let's just leave it out altogether

mpguerra:
> One thing I'm wondering about are the underlying crates with the elliptic curve implementations. In particular `ed448-goldilocks` which is not as mature as the other ones. Is this something that we're willing to include in the audit?

@chelseakomlo do you have any thoughts on this? I'd vote to leave them all out unless there's any you'd be particularly interested in getting partially audited, e.g the parts that are used by FROST code. 

Otherwise let's just close this list, merge the PR and move things forward.

dconnolly (approved):

dconnolly:
> > One thing I'm wondering about are the underlying crates with the elliptic curve implementations. In particular `ed448-goldilocks` which is not as mature as the other ones. Is this something that we're willing to include in the audit?
> 
> @chelseakomlo do you have any thoughts on this? I'd vote to leave them all out unless there's any you'd be particularly interested in getting partially audited, e.g the parts that are used by FROST code.
> 
> Otherwise let's just close this list, merge the PR and move things forward.

I think auditing each of these curve libs is a big scope increase; when we had redjubjub FROST audited they looked at the top-level library and not the `jubjub` dependency

PR #250, opened 2023-02-23
Start FROST book 

mpguerra:
I just realised that in order to add the list of dependencies for audit, we first need to create the actual FROST book using `mdbook`. This PR does that.

## Testing

Run `mdbook build --open` and check that all linked in files render properly


dconnolly (approved):
Rendered and reviewed locally, looks good!

PR #256, opened 2023-02-28
rename DKG types

### LINKED ISSUE #255: Rename DKG types
Suggestion:

`keygen_round1()` -> `round1()`  and similars

`Round1Package` -> `round1::Package` and similars

Double check if that makes sense and rename them

conradoplg:
Closes #255 

Since `dkg.rs` are very similar between crates (just the ciphersuite "letter" changes, e.g. `<E>` or `<R>`), I changed `gendoc` to use the naive search & replace to generate the whole file for all ciphersuites instead of just copying the canonical documentation (which would require refactoring the functions in all files). At this point it should be called "gencode" or something but we can do that in a separate PR.

### DIFF
--- a/frost-core/src/frost/keys/dkg.rs
+++ b/frost-core/src/frost/keys/dkg.rs
@@ -21,83 +21,93 @@ use super::{
     VerifiableSecretSharingCommitment, VerifyingShare,
 };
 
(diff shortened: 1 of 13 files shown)


dconnolly (approved):
🎉

PR #260, opened 2023-03-02
renamed gendoc to gencode; make it unpublishable

### LINKED ISSUE #257: Rename gendoc and make it non-publishable
- Rename to `gencode` or come up with a better name, since it's no longer only about documentation
- Make it unpublishable, see https://doc.rust-lang.org/cargo/reference/manifest.html#the-publish-field so that we prevent it from publishing it accidentally

conradoplg:
Closes #257


dconnolly (approved):
🎉🎉🎉

PR #274, opened 2023-03-09
Rename clippy lint derive_hash_xor_eq

dconnolly:


conradoplg:
OK that worked :thinking: The remaining failure is fixed in #277

conradoplg (approved):

Issue #227, opened 2023-02-06
Change gendoc to put comments before #[derive]s


conradoplg:
Because it was simpler, `gendoc` currently copies comments right before the function, and after `#[derive]`s.

Change it so that comments are added before `#[derive]`s, which is more usual.

PR #277, opened 2023-03-10
move comment to make gencode work

conradoplg:
`gencode` does not currently support the documentation not being directly the declaration (i.e. before `#[derive]s`).

A comment was moved in one of the files which was making  `gencode` fail. This fixes that.

The proper fix is #227

### DIFF
--- a/frost-ed25519/src/lib.rs
+++ b/frost-ed25519/src/lib.rs
@@ -19,8 +19,8 @@ mod tests;
 /// An error.
 pub type Error = frost_core::Error<Ed25519Sha512>;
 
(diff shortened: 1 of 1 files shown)


conradoplg:
I'm admin-merging to unblock other stuff, this since it only moves a comment and the CI shows it works

dconnolly:
OH this was me! I'm sorry 🫣

conradoplg:
> OH this was me! I'm sorry 🫣

No worries! I was lazy when writing gencode. Fixed it for good in #280 and now we can have comments in the right places :rofl:

PR #297, opened 2023-04-11
Re-export the frost-core traits and rand-core as part of top-level impls API

### LINKED ISSUE #296: Re-export frost-core, rand-core traits needed by consumers of top-level libraries

dconnolly:
Resolves #296

### DIFF
--- a/frost-ed25519/src/lib.rs
+++ b/frost-ed25519/src/lib.rs
@@ -11,10 +11,14 @@ use curve25519_dalek::{
 use rand_core::{CryptoRng, RngCore};
 use sha2::{Digest, Sha512};
 
-use frost_core::{frost, Ciphersuite, Field, FieldError, Group, GroupError};
+pub use frost_core::frost;
 
 #[cfg(test)]
 mod tests;
+
+// Re-exports in our public API
+pub use frost_core::{Ciphersuite, Field, FieldError, Group, GroupError};
+pub use rand_core;
 
 /// An error.
 pub type Error = frost_core::Error<Ed25519Sha512>;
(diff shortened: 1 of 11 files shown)


frost-ed25519/src/lib.rs
```diff
@@ -11,11 +11,15 @@ use curve25519_dalek::{
 use rand_core::{CryptoRng, RngCore};
 use sha2::{Digest, Sha512};
 
-use frost_core::{frost, Ciphersuite, Field, FieldError, Group, GroupError};
+pub use frost_core::frost;
```
conradoplg, line 14: This uses `pub` but others don't, was this intended?

frost-ristretto255/src/lib.rs
```diff
@@ -11,11 +11,15 @@ use curve25519_dalek::{
 use rand_core::{CryptoRng, RngCore};
 use sha2::{Digest, Sha512};
 
-use frost_core::{frost, Ciphersuite, Field, FieldError, Group, GroupError};
+pub use frost_core::frost;
```
conradoplg, line 14: This uses `pub` but others don't, was this intended?

conradoplg (changes requested):

frost-ristretto255/src/lib.rs
dconnolly: Whoops, fixed in cf6b639

frost-ed25519/src/lib.rs
dconnolly: Whoops, fixed in cf6b639

conradoplg (approved):

PR #308, opened 2023-04-18
bump version to 0.2.0, update changelog

conradoplg:
There's a breaking change so we actually need to bump to 0.2.0


dconnolly:
Should we bump the per-ciphersuite crates yet? We can leave it but

dconnolly (approved):

PR #314, opened 2023-04-21
add split_key; additional refactorings

conradoplg:
- remove `SharedSecret`, use `SigningKey` instead. The only public place it was exposed was in `reconstruct_secret`
- rename `reconstruct_secret` to `reconstruct_key`, make it return `SigningKey`, fix it to return `Error` instead of a error string
- expose `reconstruct_secret` in the ciphersuite crates
- add `split_key` (taking a `SigningKey`)


frost-core/src/frost/keys.rs
```diff
@@ -565,12 +494,23 @@ pub(crate) fn generate_secret_shares<C: Ciphersuite>(
...
+/// This can be used if for some reason the original key must be restored; e.g.
+/// if threshold signing is not required anymore.
+///
+/// This is NOT required to sign with FROST; the whole point of FROST is being
+/// able to generate signatures only using the shares, without having to
+/// reconstruct the original key.
+///
+/// The caller is responsible for providing at least `min_signers` shares;
+/// if less than that is provided, a different key will be returned.
+pub fn reconstruct_key<C: Ciphersuite>(
```
dconnolly, line 509: `reconstruct_signing_key`? or maybe even `reconstruct()`?

frost-core/src/frost/keys.rs
```diff
@@ -329,12 +242,28 @@ pub fn keygen_with_dealer<C: Ciphersuite, R: RngCore + CryptoRng>(
...
+    split_key(&key, max_signers, min_signers, rng)
+}
+
+/// Splits an existing key into FROST shares.
+///
+/// This is identical to [`keygen_with_dealer`] but receives an existing key
+/// instead of generating a fresh one. This is useful in scenarios where
+/// the key needs to be generated externally or must be derived from e.g. a
+/// seed phrase.
+pub fn split_key<C: Ciphersuite, R: RngCore + CryptoRng>(
```
dconnolly, line 256: What about just `split()`? (Non-blocking, just thinking)

frost-core/src/frost/keys.rs
```diff
@@ -600,15 +540,15 @@ pub fn reconstruct_secret<C: Ciphersuite>(
         // If at this step, the denominator is zero in the scalar field, there must be a duplicate
         // secret share.
         if den == <<C::Group as Group>::Field>::zero() {
-            return Err("Duplicate shares provided");
+            return Err(Error::DuplicatedShares);
         }
 
-        // Save numerator * 1/denomintor in the scalar field
+        // Save numerator * 1/denominator in the scalar field
```
dconnolly, line 546: 😎

frost-core/src/frost/keys.rs
```diff
@@ -565,12 +494,23 @@ pub(crate) fn generate_secret_shares<C: Ciphersuite>(
...
-pub fn reconstruct_secret<C: Ciphersuite>(
+/// Recompute the secret from at least `min_signers` secret shares
+/// using Lagrange interpolation.
+///
+/// This can be used if for some reason the original key must be restored; e.g.
+/// if threshold signing is not required anymore.
+///
+/// This is NOT required to sign with FROST; the whole point of FROST is being
+/// able to generate signatures only using the shares, without having to
+/// reconstruct the original key.
```
dconnolly, line 505:
```suggestion
/// This is NOT required to sign with FROST; the point of FROST is being
/// able to generate signatures only using the shares, without having to
/// reconstruct the original key.
```

dconnolly (commented):
looking good, some little q's about naming stuff

frost-core/src/frost/keys.rs
conradoplg: Good point. My reasoning for this name was based on the existing `keygen_with_dealer` function which also has "key" in the name. If we follow your suggestion, should we also go for `gen_with_dealer` or `generate_with_dealer`?

mpguerra:
@natalieesk do you want to have a go at reviewing this one? :)

natalieesk (changes requested):
Looks good! I think it would be good to add tests for the error cases though.

frost-core/src/frost/keys.rs
conradoplg: I went with @dconnolly 's suggestion together with mine in 5346ec2. My rationale was the Rust guideline https://rust-lang.github.io/rfcs/0356-no-module-prefixes.html. I slightly prefer the longer version like @natalieesk but it seems more idiomatic this way

frost-core/src/frost/keys.rs
conradoplg: Sounds good, done in 5346ec29

natalieesk (approved):

natalieesk (approved):
Looks good!

PR #329, opened 2023-04-26
Change gencode to handle repairable module; fix build issues

conradoplg:
This changes `gencode` to handle the repairable module: it copies the documentation from `ristretto255` into the other ciphersuite crates.

It's also possible to avoid one other redundancy and copy from `frost-core` into `ristretto255` but that will requires some other tweaks so I'll do that in a separate PR.

I also found some build issues in the ciphersuite-specific repairable modules - we forgot to declare the module so it wasn't being compiled :sweat_smile:

### DIFF
--- a/frost-ed25519/src/keys/repairable.rs
+++ b/frost-ed25519/src/keys/repairable.rs
@@ -6,7 +6,11 @@
 
 use std::collections::HashMap;
 
-use crate::{frost, Ciphersuite, CryptoRng, Identifier, RngCore, Scalar, R};
+// This is imported separately to make `gencode` work.
+// (if it were below, the position of the import would vary between ciphersuites
+//  after `cargo fmt`)
+use crate::Ed25519Sha512;
+use crate::{frost, Ciphersuite, CryptoRng, Identifier, RngCore, Scalar};
 
 use super::{SecretShare, VerifiableSecretSharingCommitment};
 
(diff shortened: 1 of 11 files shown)


frost-ed25519/src/keys/repairable.rs
```diff
@@ -1,11 +1,10 @@
-/// Repairable Threshold Schemes
-#![doc = include_str!("../../repairable.md")]
```
natalieesk, line 3: Do we not want to generate docs for this? These have been removed in this commit

natalieesk (changes requested):
The build is currently failing on formatting

conradoplg:
> The build is currently failing on formatting

Fixed it, it was an issue with  `gencode` usage (it's run in CI in the `Rustfmt` section, we probably should run it separately...)

frost-ed25519/src/keys/repairable.rs
conradoplg: I think that was confusing a comment...

After I saw your comment I copied those docs, so it should look better now

dconnolly:
> > The build is currently failing on formatting
> 
> Fixed it, it was an issue with  `gencode` usage (it's run in CI in the `Rustfmt` section, we probably should run it separately...)

Yeah it's related but not identical, right before might be good (or rustfmt before /and/ after gencode? 🤔)

dconnolly (approved):

PR #334, opened 2023-04-27
gencode repairable improvement

conradoplg:
This changes `gencode` to use the `frost-core` docs of `repairable.rs` as a base instead of `ristretto255`. This makes it a single canonical source (before you had to update both `frost-core` and `ristretto255` and run gencode to generate the rest)

This depended both on recent changes to `main` and to #329 so it was branched off #329 and synced with `main`. It needs to be rebased after #329 merges. The actual change in this PR is just the last commit so it might be simpler to cherry-pick it on top of the new `main` and force-push it in this branch.


natalieesk (approved):

natalieesk (commented):
lgtm

PR #337, opened 2023-05-02
Serde

### LINKED ISSUE #336: structs to send over the wire should implement Deserialize + Serialize
this should include at least these structs:
- dkg::round1::Package
- dkg::round2::Package
- SigningCommitments
- SignatureShare

dvc94ch:
Closes #336


conradoplg:
Thank you very much for the contribution!

However we need to consider other requirements such as making serde optional, and making sure it's the right ciphersuite when deserializing. We're figure out the design and we may use this PR as a starting point, so I'll keep it open for now.

mpguerra:
Thank you for the PR. We're going to implement this slightly differently in the end

PR #341, opened 2023-05-03
Errors caused by protocol violations should contain the misbehaving p…

dvc94ch:
…arty.


dconnolly (approved):
Looks good, thank you!

PR #361, opened 2023-05-24
Make CoefficientCommitment and VerifiableSecretSharingCommitment accessible

### LINKED ISSUE #323: Make value of VerifiableSecretSharingCommitment accessible
To access the value from VerifiableSecretSharingCommitment we need to either:

* Add a getter and a new() method
* Make the value public

natalieesk:
* Created a deserialize() and value() function for CoefficientCommitment
* Created a serialize() and deserialize() function for VerifiableSecretSharingCommitment
This closes #323


frost-core/src/frost/keys.rs
```diff
@@ -142,7 +144,27 @@ where
...
+where
+    C: Ciphersuite,
+{
+    /// Creates a new commitment from a coefficient input
+    pub fn deserialize(
+        coefficient: <C::Group as Group>::Serialization,
+    ) -> Result<CoefficientCommitment<C>, Error<C>> {
+        Ok(Self(
+            <C::Group as Group>::deserialize(&coefficient)
+                .map_err(|_| Error::from(GroupError::MalformedElement))?,
```
conradoplg, line 159: You can resurface the original error:

```suggestion
            <C::Group as Group>::deserialize(&coefficient).map_err(Error::from)?,
```

frost-core/src/frost/keys.rs
```diff
@@ -156,11 +178,58 @@ pub(super) struct CoefficientCommitment<C: Ciphersuite>(pub(super) Element<C>);
...
+    /// Returns concatenated coefficent commitments from self.0 specifying the number of commitments
+    pub fn serialize(&self) -> String {
+        let values = self
+            .0
+            .iter()
+            .map(|cc| hex::encode(<<C as Ciphersuite>::Group as Group>::serialize(&cc.0)))
+            .join("");
+
+        hex::encode(self.0.len().to_string()) + &values
+    }
```
conradoplg, line 199: We shouldn't force a particular serialization format for lists and we need to allow the caller to use their own format.

To do that we can change this to return just a vector of serialized elements, and the caller can then serialize the vector as they wish.

Sorry I didn't make this clearer! But we can still use the vector-serialization strategy you used here in the demo (where we can encode things any way we want)

```suggestion
    pub fn serialize(&self) -> Vec<<C::Group as Group>::Serialization> {
        self.0
            .iter()
            .map(|cc| <<C as Ciphersuite>::Group as Group>::serialize(&cc.0))
            .collect()
    }
```

frost-core/src/frost/keys.rs
```diff
@@ -156,11 +178,58 @@ pub(super) struct CoefficientCommitment<C: Ciphersuite>(pub(super) Element<C>);
...
+                        .map_err(|_| Error::from(GroupError::MalformedElement))?,
+                )
+                .map_err(|_| Error::InvalidCoefficient)?;
+            let coeff_commitment = CoefficientCommitment::<C>::deserialize(serialized)?;
+
+            coeff_commitments.push(coeff_commitment)
+        }
+        Ok(Self(coeff_commitments))
+    }
+}
```
conradoplg, line 231: Similarly this will require changing the function to `pub fn deserialize(coefficientCommitments: &[<C::Group as Group>::Serialization])` and changing the body accordingly (it will become much simpler). Likewise you can use the encoding logic in the demo instead.

frost-core/src/lib.rs
```diff
@@ -115,7 +115,7 @@ pub trait Group: Copy + Clone + PartialEq {
     /// A unique byte array buf of fixed length N.
     ///
     /// Little-endian!
-    type Serialization: AsRef<[u8]> + Debug + TryFrom<Vec<u8>>;
+    type Serialization: AsRef<[u8]> + Debug + TryFrom<Vec<u8>> + PartialEq;
```
conradoplg, line 118: It's better to not add an additional bound to this trait because that will require all users of the crate to implement it even if they don't want to.

In this case we can workaround by manually comparing the elements, see other suggestion

```suggestion
    type Serialization: AsRef<[u8]> + Debug + TryFrom<Vec<u8>>;
```

frost-core/src/tests.rs
```diff
@@ -325,3 +333,117 @@ where
...
+        coeff_commitments.push(coeff_commitment)
+    }
+
+    //    ---
+
+    let expected = serialize_coeff_commitments::<C>(&vec![input_1, input_2, input_3]);
+
+    let vss_commitment = VerifiableSecretSharingCommitment(coeff_commitments).serialize();
+
+    assert!(expected == vss_commitment);
```
conradoplg, line 416: See other comment. This checks if all elements in each vector are the same, with `as_ref()` since the `Serialization` type implements `AsRef<[u8]>`

```suggestion
    assert!(expected.len() == vss_commitment.len());
    assert!(expected
        .iter()
        .zip(vss_commitment.iter())
        .all(|(e, c)| e.as_ref() == c.as_ref()));
```

conradoplg (changes requested):
Thanks! I think there is just one additional change to make

frost-core/src/tests.rs
natalieesk: Done [dfc0f32](https://github.com/ZcashFoundation/frost/pull/361/commits/dfc0f324a850f164934311114eff024d2e16d993)

natalieesk:
I added a serialize() function for CoefficientCommitment as I felt that was missing. Also I did a big test refactor/rewrite.

conradoplg (approved):
:tada:

PR #397, opened 2023-06-16
remove identifiers from structs that are communicated

conradoplg:
Sending identifiers inside the structs may be problematic because it's used as the source for pinpointing cheaters when there is an error in the protocol. Even if we got rid of that, callers could mistakenly look inside the struct to find the identifier of the cheater.

This moves identifiers out of structs, and changes the API so that the caller must specify the identifier of the package being fed into the API. This will make more likely that the caller will get the identifier from the right place (from whatever mapping they have between identifiers and authenticated channels) since there is no other obvious place they could grab identifiers from.

This also makes some code a lot more straightforward, and allowed getting rid of `SignatureResponse` struct, though I had to use a workaround to encode it with serde since it's mostly an exception, and had to introduce a `SignatureShareHelper` but that's completely private and invisible to the user.


natalieesk (approved):
Looks good!

Issue #343, opened 2023-05-03
Support serde


conradoplg:
Requirements/design:

- Make it optional, behind a feature flag
- Include a ciphersuite ID when serializing and validate it when deserializing, to prevent deserializing a struct from one ciphersuite as a struct of another ciphersuite
- Do we need some versioning mechanism?

dconnolly:
We might want to make use of `serdect` to implement for our secret types in constant time:
- `SigningKey`
- `Nonce`
- `Scalar`
- `SharedSecret`
- `SigningShare`
- `SecretShare`
- ?

PR #398, opened 2023-06-20
Support serde

conradoplg:
Part of #343 

Adds optional support for serde.

This required adding a private `ciphersuite` field to each struct, which in turn made it impossible to instantiate structs directly, requiring the addition of `new()` methods. (Which seems better practice anyway).

The main approach consisted in creating structs `ElementSerialization` and `ScalarSerialization` which implement serde Serialization using the ciphersuite encoding methods, and then using `serde(into = ...)` and `serde(try_from = ...)`  to encode fields using them. This allows just `derive(Serialization, Deserialization)` in most structs and removes a lot of boilerplate code.

The `ciphersuite` field is a placeholder (does not store anything) and encoded as the ciphersuite ID. When decoding, it fails if the ciphersuite ID does not match the ciphersuite of the type being decoded.

This is still missing:

- Adding serialization support for structs like `dkg::part1::KeyPackage`. Most of the time I assume they will be just kept in memory, but it's conceivable that applications might want to make the DKG process survive the user closing an reopening the application, which would require serializing these structs.
- ~~Removing `pub` from struct fields and adding getters (the natural next step after adding new() methods; not directly related to serde supports, but seems better practice)~~ **Done in #401**
- Adding tests for other ciphersuites. We can:
  - Add a simple test for the other ciphersuites. Since the logic is generic, we can just test if the ciphersuite ID is being serialized correctly.
  - Write a generic test. We can do that by specifying a bunch of  pre-encoded scalar/elements to use as samples for the test. **This can be done easily with the sample-generating functions from #401**
- I just noticed I missed implementing Serialize/Deserialize in the other ciphersuite types (which is required by weird Rust reasons). This would be caught by the testing the other ciphersuite, so I'll do those together.
  
This PR seems big enough as is so I'll do those in separate PRs.


natalieesk (approved):
The tests are only added for ristretto here but they are added in a later PR that is also ready to be merged

PR #401, opened 2023-06-20
Remove pub from fields, add getters

conradoplg:
Depends on https://github.com/ZcashFoundation/frost/pull/398

This removes `pub` from fields and adds getters instead, which combined with the `new()` methods from #398 allow recreating structs, which in turn allows using arbitrary encoding schemes.

This also adds tests for them.

The sample-values-building functions introduced in those tests will be backported to the serde tests, allowing creating ciphersuite-generic tests.

Finally this switches to `BTreeMap` for signing commitments, which I had to touch anyway to introduce a getter for it and IMO makes the code a lot of simpler. (This was also done in #397 )


frost-ed25519/tests/recreation_tests.rs
natalieesk, line 1: I love these tests

natalieesk (changes requested):
This needs to be synced with main to avoid duplication from other PRs. Otherwise it's all good!

natalieesk (approved):
Looks good

PR #407, opened 2023-06-22
Add a comment about why we hash the message and the set of nonce comm…

dconnolly:
…itments as part of creating the preimage for the binding factor


natalieesk (approved):
Looks good

PR #409, opened 2023-06-22
expand serde tests

### LINKED ISSUE #343: Support serde

conradoplg:
Based on #401 

Closes #343 

Expands serde tests.

Also fixes serde support for other ciphersuites.

- Uses the sample-value-generating functions from #401 
- Uses gencode to replicate tests to other ciphersuites, using JSON files containing the hex encodings of the samples values in order to do the replacements.

I also considered:

- Doing the replacements at runtime inside the tests. But the tests became super ugly and required a lot of escaping inside string literals.
- Making the sample-value-generating functions also return the hex encoding of the values. But also requires doing the replacements at runtime.
- Putting the JSON literals inside files. That would make the test files smaller but I was not sure if it was worth the effort.

Feedback is welcome.


natalieesk (approved):
Looks good

PR #410, opened 2023-06-23
Prelim implementation of including the group public key in computing …

dconnolly:
…the binding factor

Needs updated test vectors from the spec PoC and updated tests

### DIFF
--- a/frost-core/src/frost.rs
+++ b/frost-core/src/frost.rs
@@ -25,7 +25,8 @@ pub mod round1;
 pub mod round2;
 
 use crate::{
-    scalar_mul::VartimeMultiscalarMul, Ciphersuite, Element, Error, Field, Group, Scalar, Signature,
+    scalar_mul::VartimeMultiscalarMul, Ciphersuite, Element, Error, Field, Group, Scalar,
+    Signature, VerifyingKey,
 };
 
 pub use self::identifier::Identifier;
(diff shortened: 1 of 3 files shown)


frost-core/src/frost.rs
```diff
@@ -238,10 +240,19 @@ where
...
     ) -> Vec<(Identifier<C>, Vec<u8>)> {
         let mut binding_factor_input_prefix = vec![];
 
+        // The length of a serialized verifying key of the same cipersuite does
+        // not change between runs of the protocol, so we don't need to hash to
+        // get a fixed length.
+        //
+        // TODO: when serde serialization merges, change this to be simpler?
+        binding_factor_input_prefix
+            .extend_from_slice(<C::Group>::serialize(&group_public.element).as_ref());
```
conradoplg, line 254: I don't think serde helps here, it helps more for serializing structs. But you can use `group_public.to_bytes()` which should simplify it.

dconnolly: yeah that's much better

conradoplg:
I incorporated the test vector format update from https://github.com/ZcashFoundation/frost/pull/430, copied the new test vectors and generated vectors for "big identifier" tests.

It seems now good to go

natalieesk (changes requested):
I've reviewed the changes from Conrado since his review. All looks good. This has some conflicts with main now that need to be resolved and then good to go.

natalieesk (approved):

natalieesk (approved):
Great

edwin1729:
Hello @natalieesk @dconnolly @conradoplg. Do you have a simple way of generating the test vector files `vectors.json`, etc. I am implementing a new frost ciphersuite (for the mina blockchain), and if you have an easy of generating these json files, that would be great for me to use. Thanks

conradoplg:
> Hello @natalieesk @dconnolly @conradoplg. Do you have a simple way of generating the test vector files `vectors.json`, etc. I am implementing a new frost ciphersuite (for the mina blockchain), and if you have an easy of generating these json files, that would be great for me to use. Thanks

They were generated with [the reference implementation](https://github.com/cfrg/draft-irtf-cfrg-frost/tree/master/poc), so unfortunately that won't be of much use to you unless you implement the ciphersuite in Python

PR #418, opened 2023-06-27
add support for deriving identifiers from arbitary strings

### LINKED ISSUE #376: Allow deriving identifiers from arbitrary strings
To make it easier to use things like emails or usernames as identifiers, we could provide a `&[u8]` -> `Identifier` map with a hash-to-scalar function (which we already have, though we would need Yet Another Hash Function)

conradoplg:
Closes #376 

This also cleans up a test which had the wrong name (`frost-ed25519/tests/integration_test.rs` should be `_tests`) and moves the unique part of it (the interoperability test) to a separate file. This allows adding `integrations_tests` to `gencode`.


frost-p256/src/lib.rs
```diff
@@ -225,6 +225,14 @@ impl Ciphersuite for P256Sha256 {
             m,
         ))
     }
+
+    /// HID for FROST(P-256, SHA-256)
+    fn HID(m: &[u8]) -> Option<<<Self::Group as Group>::Field as Field>::Scalar> {
```
natalieesk, line 230: This function is different than in the other ciphersuites but does the same thing. I think it would be good to have them all be written the same way. (Unless I'm misunderstaind something)

natalieesk (changes requested):
Looks good other than a comment about making the HID functions match each other across ciphersuites

conradoplg (commented):
Synced with main and answered comment

natalieesk (approved):

PR #419, opened 2023-06-27
allow specifying identifiers in generate_with_dealer

### LINKED ISSUE #375: Allow applications to specify which identifier to use
With trusted dealer the library currents forces the usage of incremental identifiers in the `[1, n]` range, which may be cumbersome if the application already has some form of identifier for participants.

Allow the caller to specify which identifier to use during key generation.

conradoplg:
Closes #375 

I'm not sure about the API, feedback is welcome. Possible alternatives:

- Create a new `generate_with_dealer_for_identifiers` (too long?)
- Just pass the list instead of enum, and auto-generate if empty (not very idiomatic...)


natalieesk:
I think passing an enum is fine as it's much cleaner. And then we can have one function. In the Rust documentation it describes a catch all pattern so I think it is idiomatic.

natalieesk (commented):
I'm happy to approve this if you agree about keeping the enum

conradoplg:
> I'm happy to approve this if you agree about keeping the enum

Sounds good! I also synced with main

conradoplg:
synced again

natalieesk (approved):

Issue #374, opened 2023-06-03
Simplify extracting identifier from cheaters


conradoplg:
Some error types contain identifiers which indicate which participant misbehaved in the protocol. However, extracting this information is awkward for callers (they need to `match` the error and get the identifier for each type that contains them).

Make this easier. Possible approaches:

- Add a `identifier()` or `participant()` method to  `Error`
- Some trait magic?

PR #420, opened 2023-06-28
add Error::culprit() to make it easier to find the identifier of misbehaving participants

conradoplg:
Closes https://github.com/ZcashFoundation/frost/issues/374

I'm not sure about the method name, suggestions are welcome...

This also renames some of the error fields for consistency, and removes the identifier from `InvalidSecretShare` because I realized it doesn't make sense in this case. If the secret share is invalid, then it's the Coordinator's fault (and not fault of the person who received the secret share!).

I  added the test just to `ristretto255`; it will eventually be replicated to other ciphersuites after #418 merges.

### DIFF
--- a/frost-core/src/error.rs
+++ b/frost-core/src/error.rs
@@ -4,6 +4,9 @@ use thiserror::Error;
 
 use crate::{frost::Identifier, Ciphersuite};
 
+#[derive(Error, Debug, Clone, Copy, Eq, PartialEq)]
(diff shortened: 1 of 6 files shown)


natalieesk:
Agree on the `InvalidSecretShare` point. Also from a demo point of view, if an individual participant has an invalid share it doesn't make much sense to have to return the error with their identifier.

natalieesk:
I think culprit() is a bit intense but I also can't think of anything better!

natalieesk (approved):
This looks good

Issue #417, opened 2023-06-27
Implement Debug and common traits


conradoplg:
It's good practice to [implement common traits](https://rust-lang.github.io/api-guidelines/interoperability.html#types-eagerly-implement-common-traits-c-common-traits).

Do that for our public types.

Make sure Debug does not print secret values.

PR #422, opened 2023-06-28
implement common traits for public types

conradoplg:
Closes https://github.com/ZcashFoundation/frost/issues/417


natalieesk (changes requested):
This looks good but I think there may be tests missing for Debug for CoefficientCommitment and VerifiableSecretSharingCommitment

conradoplg:
> This looks good but I think there may be tests missing for Debug for CoefficientCommitment and VerifiableSecretSharingCommitment

Thanks!
These are being tested indirectly since they are included in other structs that are tested, and we are mostly using `derive(Debug)` (which requires the fields to also implement `Debug`)

natalieesk (approved):

PR #435, opened 2023-07-10
fix curve name in frost-secp256k1 crate description

dufkan:
The description of the `frost-secp256k1` crate currently states that it uses the Ristretto group.


conradoplg (approved):
Thanks!

PR #436, opened 2023-07-10
refactor Lagrange coefficient computation

### LINKED ISSUE #320: Refactor `compute_lagrange_coefficient`
To remove nearly duplicated code. See https://github.com/ZcashFoundation/frost/pull/281#discussion_r1136117403

conradoplg:
This helps with the simplified rerandomized FROST, so I did it in preparation for that.

I think this surfaced some error cases that weren't being handled correctly before (i.e. we weren't checking if the given identifier was included in the set).

Closes #320


frost-core/src/frost.rs
```diff
@@ -148,34 +148,67 @@ where
...
+/// The Lagrange polynomial for a set of points (x_j, y_j) for 0 <= j <= k
+/// is ∑_{i=0}^k y_i.ℓ_i(x), where ℓ_i(x) is the Lagrange basis polynomial:
+///
+/// ℓ_i(x) = ∏_{0≤j≤k; j≠i} (x - x_j) / (x_i - x_j).
+///
+/// This computes ℓ_j(x) for the set of points `xs` and for the j corresponding
+/// to the given xj.
+///
+/// If `x` is None, it uses 0 for it (since Identifiers can't be 0)
+pub fn compute_lagrange_coefficient<C: Ciphersuite>(
```
dconnolly, line 162: <3

dconnolly: I'm torn between these unicode chars that are good and doing this with LaTeX which we can render nicely in the rustdoc...

dconnolly: Ah vanilla, yes, but we can pull in katex/mathjax in the html include in the header, which is a stable feature now, and docs.rs will render it, but it's a pro/con on the actual rustdoc text readability when inline with code

frost-core/src/frost.rs
```diff
@@ -148,34 +148,67 @@ where
...
+    xs: &BTreeSet<Identifier<C>>,
+    x: Option<Identifier<C>>,
```
dconnolly, line 164: Before i start suggesting renames, what about `x_set` instead of `xs`, and `x_i` instead of `xi`?

dconnolly: > Before i start suggesting renames, what about `x_set` instead of `xs`, and `x_i` instead of `xi`? (and then inside use `x_j`)

frost-core/src/frost/keys.rs
```diff
@@ -754,39 +756,20 @@ pub fn reconstruct<C: Ciphersuite>(
 
     let mut secret = <<C::Group as Group>::Field>::zero();
 
-    // Compute the Lagrange coefficients
-    for (i_idx, i, secret_share) in secret_shares
+    let xset: BTreeSet<_> = secret_shares
```
dconnolly, line 759: aha, any problem using similar variable names out here and inside `compute_lagrange_coefficients()`?

frost-core/src/frost.rs
```diff
@@ -148,34 +148,67 @@ where
...
         }
 
-        num *= *commitment_identifier;
-
-        den *= *commitment_identifier - *signer_id;
+        if let Some(x) = x {
+            num *= x - *xj;
+            den *= xi - *xj;
+        } else {
+            // Both signs inverted just to avoid requiring Neg (-*xj)
```
dconnolly, line 184: 👍

frost-core/src/frost.rs
```diff
@@ -148,34 +148,67 @@ where
...
+/// If `x` is None, it uses 0 for it (since Identifiers can't be 0)
+pub fn compute_lagrange_coefficient<C: Ciphersuite>(
+    xs: &BTreeSet<Identifier<C>>,
+    x: Option<Identifier<C>>,
+    xi: Identifier<C>,
 ) -> Result<Scalar<C>, Error<C>> {
-    let zero = <<C::Group as Group>::Field>::zero();
-
+    if xs.is_empty() {
+        return Err(Error::IncorrectNumberOfIdentifiers);
```
dconnolly, line 168: 👍

frost-core/src/frost.rs
```diff
@@ -148,34 +148,67 @@ where
...
+    }
     let mut num = <<C::Group as Group>::Field>::one();
     let mut den = <<C::Group as Group>::Field>::one();
 
-    for commitment_identifier in signing_package.signing_commitments().keys() {
-        if *commitment_identifier == *signer_id {
+    let mut xi_found = false;
+    for xj in xs.iter() {
```
dconnolly, line 174: little style nit

```suggestion

    for xj in xs.iter() {
```

frost-core/src/frost.rs
```diff
@@ -148,34 +148,67 @@ where
...
+        } else {
+            // Both signs inverted just to avoid requiring Neg (-*xj)
+            num *= *xj;
+            den *= *xj - xi;
+        }
     }
-
-    if den == zero {
-        return Err(Error::DuplicatedShares);
+    if !xi_found {
```
dconnolly, line 189: Should we return this immediately after the `xi_found` loop above?

frost-core/src/frost.rs
dconnolly: (this is non-blocking, just a continual thing I go back and forth on with math in rustdoc :D)

frost-core/src/frost.rs
```diff
@@ -148,34 +148,68 @@ where
...
+        return Err(Error::UnknownIdentifier);
     }
 
-    let lagrange_coeff = num * <<C::Group as Group>::Field>::invert(&den).unwrap();
+    Ok(num
+        * <<C::Group as Group>::Field>::invert(&den).map_err(|_| Error::DuplicatedIdentifiers)?)
+}
 
-    Ok(lagrange_coeff)
+/// Generates the lagrange coefficient for the i'th participant.
```
dconnolly, line 198:
```suggestion
/// Generates the lagrange coefficient for the i'th participant (for `signer_id`).
```

frost-core/src/frost.rs
```diff
@@ -148,34 +148,68 @@ where
...
-        num *= *commitment_identifier;
-
-        den *= *commitment_identifier - *signer_id;
+        if let Some(x) = x {
+            num *= x - *xj;
+            den *= xi - *xj;
+        } else {
+            // Both signs inverted just to avoid requiring Neg (-*xj)
+            num *= *xj;
+            den *= *xj - xi;
```
dconnolly, line 187: I think this is self-consistent when applying the `∏`? I guess it will always be `Some(x)` or always `x = 0`, so

dconnolly (commented):
Some mostly small q's before approval, looks nice

frost-core/src/frost.rs
conradoplg: Ah got it, I wasn't aware `--html-in-header` was supported in docs.rs. I'm inclined to leave as is since the formulas aren't that complicated, but I can change it in a followup PR if you prefer to

frost-core/src/frost.rs
conradoplg: That's indeed much better, changed in a3558ea

frost-core/src/frost.rs
conradoplg: Sorry, I don't get what you mean

frost-core/src/frost.rs
conradoplg: I think we can't, we are determining if x_i is inside the set or not. We'll only be sure after we iterate the entire set.

I just realized this is a microoptimization though, we can just check if `!x_set.contains(&x_i)` before the loop if you prefer

frost-core/src/frost.rs
conradoplg: Also this made me realize the function is `pub` but it doesn't need to be, I made it optionally pub under `internals` feature in a3558ea

frost-core/src/frost/keys.rs
conradoplg: Good point. I went for `identifiers` in a3558ea which seems more semantic, what do you think?

dconnolly (approved):
♥️

PR #438, opened 2023-07-11
Update test vectors

conradoplg:
Updated to the (hopefully final) test vectors from https://www.ietf.org/archive/id/draft-irtf-cfrg-frost-14.html

Also updated the spec links, though we'll need to update them again after the spec is finalized.


natalieesk (approved):
lgtm

PR #453, opened 2023-07-26
Add minimum participant constraints

### LINKED ISSUE #443: [NCC-E008263-XLV] `frost-core`: Minimum Participant Constraint Enforcement Improvements
`generate_secret_polynomial` contains common input validation for min/max_signers, but these validations happen too late. In particular, `split()` can panic since it passes `min_signers -1`.

- Refactor the validations into a separate function if possible, and call it where appropriate.
- Fix the panic in `split()`

One idea is to create a type e.g. "ValidConfig" which can only be constructed with a valid configuration and use that everywhere that is needed.

natalieesk:
This closes #443


conradoplg (commented):
Looks good! Do you want to include the new findings in this PR too?

- DKG `part1()`
- `repair_share_step_1()`

conradoplg (approved):
Looks great, thanks!

Issue #455, opened 2023-07-26
[NCC-E008263-AW3] Insufficient Participant Commitment List Checks


conradoplg:
`sign()` does not properly validate the `signing_package.signing_commitments`. See report for more details.
 
 Fix it, so it checks:

- If the participant's identifier and commitment is included (done in https://github.com/ZcashFoundation/frost/pull/452)
- If there aren't any repeated commitments
- Check consistency with min_signers/max_signers? We'll need to think a bit about this since those aren't readily available. It may be a good idea to store min_signers in the KeyPackage, but max_signers might not possible (since it's possible to issue new shares).

PR #480, opened 2023-08-17
check number of commitments in sign()

conradoplg:
Closes https://github.com/ZcashFoundation/frost/issues/455

Not fully sure if it's worth checking this (it's not required, the signature generation would fail anyway), but it works as a second layer of validation and it seems useful for users to be able to know what is the min_signers value by getting it from the KeyPackage. Let me know what you think.

This also adds a test for trying to sign with less than min_signers share (even bypassing this new min_signers value check) which is useful in itself and makes sure we're not interpreting thresholds wrong.

### DIFF
--- a/frost-core/src/tests/ciphersuite_generic.rs
+++ b/frost-core/src/tests/ciphersuite_generic.rs
@@ -1,4 +1,6 @@
 //! Ciphersuite-generic test functions.
+#![allow(clippy::type_complexity)]
+
 use std::{
     collections::{BTreeMap, HashMap},
     convert::TryFrom,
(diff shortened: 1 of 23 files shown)


conradoplg:
Solved conflicts

frost-core/src/tests/ciphersuite_generic.rs
```diff
@@ -117,8 +119,24 @@ pub fn check_sign_with_dealer<C: Ciphersuite, R: RngCore + CryptoRng>(
         let key_package = frost::keys::KeyPackage::try_from(v).unwrap();
         key_packages.insert(k, key_package);
     }
+    // Check if it fails with not enough signers. To bypass the min_signers value
```
natalieesk, line 122: I find this comment a bit confusing

frost-core/src/tests/ciphersuite_generic.rs
```diff
@@ -117,8 +119,24 @@ pub fn check_sign_with_dealer<C: Ciphersuite, R: RngCore + CryptoRng>(
...
+    // check, also reduce the min_signers in KeyPackages.
```
natalieesk, line 123: I find this comment a bit confusing

natalieesk (changes requested):
Looks good. Just one thing about making a comment clearer.

frost-core/src/tests/ciphersuite_generic.rs
conradoplg: I expanded it in fa7f652

natalieesk (approved):

natalieesk (dismissed):

Issue #456, opened 2023-07-26
[NCC-E008263-9XW] Missing Length Check in Identifiers List


conradoplg:
`split()` is not validating if the identifier list provided, if any, has the correct length, which can lead to multiple problems. See report for more details.

Check if it matches `max_signers` and return error otherwise.

Issue #460, opened 2023-07-26
Generate default identifiers only when needed


conradoplg:
In `split()` we always generate the default list of identifiers, but we could generate them only when needed.

I think I did that due to lifetime reasons in Rust, but double check if we can improve that.

conradoplg:
Done in https://github.com/ZcashFoundation/frost/pull/481

PR #481, opened 2023-08-17
check if the correct number of identifiers was provided in split()

conradoplg:
Closes https://github.com/ZcashFoundation/frost/issues/456 and https://github.com/ZcashFoundation/frost/issues/460

### DIFF
--- a/frost-core/src/frost/keys.rs
+++ b/frost-core/src/frost/keys.rs
@@ -474,18 +474,25 @@ pub fn split<C: Ciphersuite, R: RngCore + CryptoRng>(
 ) -> Result<(HashMap<Identifier<C>, SecretShare<C>>, PublicKeyPackage<C>), Error<C>> {
     validate_num_of_signers(min_signers, max_signers)?;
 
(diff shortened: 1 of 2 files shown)


frost-core/src/frost/keys.rs
```diff
@@ -474,18 +474,25 @@ pub fn split<C: Ciphersuite, R: RngCore + CryptoRng>(
...
-        IdentifierList::Custom(identifiers) => identifiers,
-        IdentifierList::Default => &default_identifiers,
+    let secret_shares = match identifiers {
+        IdentifierList::Default => {
+            let identifiers = default_identifiers(max_signers);
+            generate_secret_shares(key, max_signers, min_signers, coefficients, &identifiers)?
+        }
+        IdentifierList::Custom(identifiers) => {
+            generate_secret_shares(key, max_signers, min_signers, coefficients, identifiers)?
+        }
```
conradoplg, line 494: This is a bit ugly but it was the cleanest way to avoid allocating the default identifiers when not needed.

dconnolly (approved):
👍👍

Issue #458, opened 2023-07-26
validate length of shares in reconstruct()


conradoplg:
`reconstruct()` does not validate if the length of the shares is >= min_shares. See audit report for details.

There is no obvious way to get the min_shares from since the participant since they lost their share and are reconstructing it, so we might want just to add it as parameter for best-effort checking purposes.

One idea is to create wrapper types for the values being sent during share recovery, rather just sending scalars back and forth. Then the other participants can send min_signers along with the shares and the participant can check if they are all equal and use that.

PR #482, opened 2023-08-17
change reconstruct() to take KeyPackages; validate size

conradoplg:
Depends on https://github.com/ZcashFoundation/frost/pull/480

Closes https://github.com/ZcashFoundation/frost/issues/458

This also changes `reconstruct()` to take `KeyPackages`s instead of `SecretShare`s which made no sense - users are expect to convert `SecretShare`s into `KeyPackage`s and store that.


natalieesk (approved):
Looks good

Issue #465, opened 2023-07-27
Improve handling of identity/zero in VerifyingKey/SigningKey


conradoplg:
While FROST does not allow the identity, the implementation allows creating a VerifyingKey with the identity or a SigningKey with a zero scalar.

Check if this can be improved and if it's OK to not allow these values. In particular, the RedDSA ciphersuites may need to allow these for compatibility with the RedDSA definition, so the solution might need to be ciphersuite-specific.

conradoplg:
FROST users will probably never instantiate `SigningKey` directly so it seems fine blocking zero anyway. I don't see any issue with RedDSA compatibility, since RedDSA users would use `reddsa` directly (and FROST RedDSA users also don't need to create `SigningKey` directly)

PR #496, opened 2023-08-22
return error when creating a zero SigningKey

conradoplg:
Closes https://github.com/ZcashFoundation/frost/issues/465

Since `VerifyingKey::new()` is private, it shouldn't possible to create one from identity, so this addresses the whole issue.

### DIFF
--- a/frost-core/src/signing_key.rs
+++ b/frost-core/src/signing_key.rs
@@ -30,6 +30,7 @@ where
     ) -> Result<SigningKey<C>, Error<C>> {
         let scalar =
             <<C::Group as Group>::Field as Field>::deserialize(&bytes).map_err(Error::from)?;
+
         if scalar == <<C::Group as Group>::Field as Field>::zero() {
             return Err(Error::MalformedSigningKey);
         }
--- a/frost-core/src/signing_key.rs
+++ b/frost-core/src/signing_key.rs
@@ -34,6 +34,7 @@ where
         if scalar == <<C::Group as Group>::Field as Field>::zero() {
             return Err(Error::MalformedSigningKey);
         }
+
         Ok(Self { scalar })
     }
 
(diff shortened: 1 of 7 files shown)


frost-core/src/signing_key.rs
```diff
@@ -28,9 +28,12 @@ where
     pub fn deserialize(
         bytes: <<C::Group as Group>::Field as Field>::Serialization,
     ) -> Result<SigningKey<C>, Error<C>> {
-        <<C::Group as Group>::Field as Field>::deserialize(&bytes)
-            .map(|scalar| SigningKey { scalar })
-            .map_err(|e| e.into())
+        let scalar =
+            <<C::Group as Group>::Field as Field>::deserialize(&bytes).map_err(Error::from)?;
+        if scalar == <<C::Group as Group>::Field as Field>::zero() {
```
dconnolly, line 33:
```suggestion

        if scalar == <<C::Group as Group>::Field as Field>::zero() {
```

frost-core/src/signing_key.rs
```diff
@@ -28,9 +28,13 @@ where
...
-        <<C::Group as Group>::Field as Field>::deserialize(&bytes)
-            .map(|scalar| SigningKey { scalar })
-            .map_err(|e| e.into())
+        let scalar =
+            <<C::Group as Group>::Field as Field>::deserialize(&bytes).map_err(Error::from)?;
+
+        if scalar == <<C::Group as Group>::Field as Field>::zero() {
+            return Err(Error::MalformedSigningKey);
+        }
+        Ok(Self { scalar })
```
dconnolly, line 37:
```suggestion

        Ok(Self { scalar })
```

dconnolly (approved):
💪

PR #502, opened 2023-08-28
Refactor computing the public key package and expose it.

dvc94ch:
These are useful for other crates. Our particular use case is integrating frost with substrate. So we store the verifiable secret sharing commitments on chain. If a node crashes it can recover and recompute the public key package from the on chain commitments.

### DIFF
--- a/frost-core/src/frost/keys.rs
+++ b/frost-core/src/frost/keys.rs
@@ -50,7 +50,7 @@ pub(crate) fn compute_verifying_share<C: Ciphersuite>(
     group_commitment: &VerifiableSecretSharingCommitment<C>,
     peer: Identifier<C>,
 ) -> VerifyingShare<C> {
-    VerifyingShare::new(evaluate_vss(group_commitment, peer))
+    VerifyingShare(evaluate_vss(group_commitment, peer))
 }
 
 /// Computes the group public key given the group commitment.
@@ -58,7 +58,9 @@ pub(crate) fn compute_verifying_share<C: Ciphersuite>(
 pub(crate) fn compute_public_key<C: Ciphersuite>(
     group_commitment: &VerifiableSecretSharingCommitment<C>,
 ) -> VerifyingKey<C> {
-    VerifyingKey::new(group_commitment.first().expect("valid commitments").value())
+    VerifyingKey {
+        element: group_commitment.coefficients()[0].value(),
+    }
 }
 
 /// Computes the public key package given a list of commitments.
--- a/frost-core/src/frost/keys.rs
+++ b/frost-core/src/frost/keys.rs
@@ -44,38 +44,6 @@ pub(crate) fn compute_group_commitment<C: Ciphersuite>(
     VerifiableSecretSharingCommitment(group_commitment)
 }
 
-/// Computes a verifying share for a peer given the group commitment.
-#[cfg_attr(feature = "internals", visibility::make(pub))]
-pub(crate) fn compute_verifying_share<C: Ciphersuite>(
-    group_commitment: &VerifiableSecretSharingCommitment<C>,
-    peer: Identifier<C>,
-) -> VerifyingShare<C> {
-    VerifyingShare(evaluate_vss(group_commitment, peer))
-}
-
-/// Computes the group public key given the group commitment.
-#[cfg_attr(feature = "internals", visibility::make(pub))]
-pub(crate) fn compute_public_key<C: Ciphersuite>(
-    group_commitment: &VerifiableSecretSharingCommitment<C>,
-) -> VerifyingKey<C> {
-    VerifyingKey {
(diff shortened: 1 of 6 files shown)


conradoplg:
Don't worry about clippy, we'll fix these shortly, some news errors came up in the Rust release last week

conradoplg:
We have just fixed the clippy errors on main, if you rebase on top of it it should pass CI

dvc94ch:
Done

dvc94ch:
Kind of annoying that `generate_trusted_dealer` uses a different method of computing the public key than `dkg`. Was very confusing, was going insane trying to figure out how my implementation was not equivalent to the previous one. Obviously it is, it's just that `generated_trusted_dealer` only generates one set of commitments. Seems like the commitments need to be divided by max_signers so that summing them produces the original commitment. However that would be non compliant with the spec appendix. Any thoughts on what to do about it?

frost-core/src/frost/keys.rs
```diff
@@ -28,6 +28,52 @@ use super::compute_lagrange_coefficient;
...
+    members: &BTreeSet<Identifier<C>>,
+    group_commitment: &VerifiableSecretSharingCommitment<C>,
+) -> PublicKeyPackage<C> {
+    let mut verifying_keys = HashMap::new();
+    for peer in members {
+        verifying_keys.insert(*peer, compute_verifying_share(group_commitment, *peer));
+    }
+    PublicKeyPackage::new(verifying_keys, compute_public_key(group_commitment))
+}
+
```
dvc94ch, line 76: Now that I understand it better the better api would probably be to implement `Add` for `VerifiableSecretSharingCommitment` and add `from_commitment` methods to the `VerifyingShare`, `VerifyingKey` and `PublicKeyPackage`.

frost-core/src/frost/keys/dkg.rs
```diff
@@ -485,44 +443,26 @@ pub fn part3<C: Ciphersuite>(
...
-    // Add the participant's own public verification share for consistency
-    all_verifying_keys.insert(round2_secret_package.identifier, verifying_key);
-
+    let members = round1_packages.keys().copied().collect();
+    let commitments: Vec<_> = round1_packages
+        .values()
+        .map(|package| package.commitment.clone())
+        .chain(iter::once(round2_secret_package.commitment.clone()))
+        .collect();
+    let group_commitment = compute_group_commitment(&commitments);
```
dvc94ch, line 456:
```suggestion
    let commitment = round1_packages
        .values()
        .map(|package| package.commitment.clone())
        .chain(iter::once(round2_secret_package.commitment.clone()))
        .reduce(|acc, e| acc + e);
```

conradoplg:
> Kind of annoying that `generate_trusted_dealer` uses a different method of computing the public key than `dkg`. Was very confusing, was going insane trying to figure out how my implementation was not equivalent to the previous one. Obviously it is, it's just that `generated_trusted_dealer` only generates one set of commitments. Seems like the commitments need to be divided by max_signers so that summing them produces the original commitment. However that would be non compliant with the spec appendix. Any thoughts on what to do about it?

The `VerifiableSecretSharingCommitment` in trusted dealer keygen and in DKG are semantically different. In trusted dealer all participants receive the same VSSCommitment, which commits to the polynomial generated by the trusted dealer. In DKG each participant has its own value, which commits to the polynomial each of them have generated.

Thus this PR will need two methods: one for deriving a PublicKeyPackage from a single VSSCommitment, to use in trusted dealer, and one for deriving a PublicKeyPackage from a list of VSSCommitments, to use in DKG.

We're currently doing a bunch of refactorings which will conflict with this PR, so if you prefer I can do both the adjustment I mentioned and the rebasing on top of the refactorings.

dvc94ch:
Yes, I understand that now. But PublicKeyPackage doesn't need two methods, just one. In dkg you can sum all the commitments to a single one which can be used to produce a PublicKeyPackage. This is what I called a group commitment. Not sure if there is a better word for it.

dvc94ch:
So for a dkg you'd call `PublicKeyPackage::from_commitment(compute_group_commitment(&commitments))`

conradoplg:
Good point! But I think it may be confusing to users. Also indeed the name `group_commitment` conflicts with another `group_commitment` in the protocol (there is even already a `compute_group_commitment` in the spec). I'd go with `sum_commitments` but I think it's better to have two different methods. In any case I'll discuss this with the team.

conradoplg:
Thank you for this, indeed it makes a lot of sense and it helps organizing things.

I'm closing this in favor of https://github.com/ZcashFoundation/frost/pull/551, because I made a mess with the history after syncing with main and I didn't want to force push your branch, so I created that instead based on your commits. I made some changes to improve consistency with the rest of the code base but the API should be equivalent.

Issue #504, opened 2023-08-29
Check that the set of identifiers between DKG rounds is consistent


conradoplg:
We should check whether the set of identifiers in `part3()` is the same as the one in `part2()`, by storing the set in the `SecretPackage` and comparing it.

PR #506, opened 2023-08-30
check if the part3() identifier consistency check works

conradoplg:
Closes https://github.com/ZcashFoundation/frost/issues/504

Turns out I already had implemented the check :facepalm: but I did write a test for it in this PR


frost-core/src/tests/ciphersuite_generic.rs
```diff
@@ -449,6 +456,43 @@ where
     check_sign(min_signers, key_packages, rng, pubkeys)
 }
 
+/// Check that calling dkg::part3() with distinct sets of participants fail.
```
dconnolly, line 459: 👍

dconnolly (approved):
nice!

PR #508, opened 2023-08-30
simplify ciphersuite encoding, without changing format

conradoplg:
While preparing for a possible default serialization format, I wanted to clean up some technical debt and reduce boilerplate code for serializing the ciphersuite.

This does not change the format, but does now allow extra fields in self-describing formats (e.g. JSON) because `serde(flatten)` (which is required for the PR) doesn't work with `deny_unknown_fields`. I think this is OK since I don't see how extra fields can be a problem.


conradoplg:
`flatten` is not compatible with `postcard` so we'll have to close this :smiling_face_with_tear:

PR #509, opened 2023-08-30
ci: run tests with --release

conradoplg:
Tests are much faster with `--release` so this adds the flag in CI


conradoplg:
It's 3 min vs 5 min, but since coverage takes much longer it doesn't make much difference, feel free to approve or close this

dconnolly (approved):

PR #491, opened 2023-08-19
improve book per audit report

conradoplg:
Closes https://github.com/ZcashFoundation/frost/issues/466

I split the tutorial into multiple subsections to make the structure less confusing. This will make the edits harder to see - sorry. Try using `git diff --color-moved origin/main`.

PR #511, opened 2023-09-01
Add default serialization format

conradoplg:
Depends on #491 due to documentation.

This adds postcard-serde-encoded serialization as the default, via `(de)serialize()` methods to encapsulate serde&postcard so that users don't need to interact with them (and we can drop serde at some point if we wish to).

I still want to add a version field in the beginning, and move the ciphersuite ID after it, but I'll do that in a separate PR.


conradoplg:
solved conflicts

conradoplg:
Rebased after base PR merged

conradoplg:
Solved conflicts

book/src/user/serialization.md
```diff
@@ -0,0 +1,60 @@
...
+- Integers are encoded in [varint
+  format](https://postcard.jamesmunns.com/wire-format#varint-encoded-integers)
+- Fixed-size byte arrays are encoded as-is (e.g. scalars, elements)
+  - Note that the encoding of scalars and elements are defined by the
+    ciphersuites.
+- Variable-size byte arrays are encoded with a length prefix (varint-encoded)
+  and the array as-is (e.g. the message)
+- Maps are encoded as the varint-encoded item count, followed by concatenated
+  item encodings.
+- Ciphersuite IDs are encoded as the 4-byte CRC-32 of the ID string.
```
dconnolly, line 18: The ID string or the Ciphersuite string? Or 'yes'?

conradoplg: Ciphersuite ID string? i.e. `FROST(Ed25519, SHA-512)`. I can see it's confusing, not sure what to call it.

book/src/user/serialization.md
```diff
@@ -0,0 +1,60 @@
...
+The format is basically the `serde` encoding of the structs using the
+[`postcard`](https://docs.rs/postcard/latest/postcard/) crate.
+
+- Integers are encoded in [varint
+  format](https://postcard.jamesmunns.com/wire-format#varint-encoded-integers)
+- Fixed-size byte arrays are encoded as-is (e.g. scalars, elements)
+  - Note that the encoding of scalars and elements are defined by the
+    ciphersuites.
+- Variable-size byte arrays are encoded with a length prefix (varint-encoded)
+  and the array as-is (e.g. the message)
```
dconnolly, line 15: 👍

book/src/user/serialization.md
```diff
@@ -0,0 +1,60 @@
...
+- Maps are encoded as the varint-encoded item count, followed by concatenated
+  item encodings.
```
dconnolly, line 17: 👍

book/src/user/serialization.md
dconnolly, line 18:
```suggestion
- Ciphersuite IDs are encoded as the 4-byte CRC-32 of the ciphersuite ID string (not the ciphersuite contextString from the FROST specification).
```

dconnolly, line 18: If the IDs are encoded as the 4-byte CRC-32 of a string, does that make it more fine to lean on the spec contextString parameter instead?

book/src/user/serialization.md
```diff
@@ -0,0 +1,60 @@
...
+- `e6811b69`: the ciphersuite ID of the SigningCommitments, CRC-32 of "FROST(ristretto255, SHA-512)"
+- `0b`: the length of the message
+- `68656c6c6f20776f726c64`: the message
+- `e6811b69`: the ciphersuite ID of the SigningPackage, CRC-32 of "FROST(ristretto255, SHA-512)"
+
+```admonish note
+The ciphersuite ID is encoded multiple times in this case because `SigningPackage` includes
+`SigningCommitments`, which also need to be communicated in Round 1 and thus also encodes
+its ciphersuite ID. This is the only instance where this happens.
+```
```
dconnolly, line 54: 👍

frost-core/src/lib.rs
```diff
@@ -243,6 +255,10 @@ pub trait Ciphersuite: Copy + Clone + PartialEq + Debug {
     /// The ciphersuite ID string
     const ID: &'static str;
 
+    /// The short 4-byte ID. Automatically derived as the CRC-32 of the UTF-8
+    /// encoded ID in big endian format.
+    const SHORT_ID: u32 = const_crc32::crc32(Self::ID.as_bytes());
```
dconnolly, line 260: 👍

dconnolly: Actually, does this need to be part of this trait, when we can directly derive it as-is? We are still in beta so it's ok the change the trait but figure I'd ask

frost-core/src/lib.rs
```diff
@@ -422,10 +444,55 @@ where
...
+    }
+}
+
+// Default byte-oriented serialization for structs that need to be communicated.
+//
+// Note that we still manually implement these methods in each applicable type,
+// instead of making these traits `pub` and asking users to import the traits.
+// The reason is that ciphersuite traits would need to re-export these traits,
+// parametrized with the ciphersuite, but trait aliases are not currently
+// supported: <https://github.com/rust-lang/rust/issues/41517>
```
dconnolly, line 470: 👍

dconnolly (commented):
Looking good, just some specific questions on the associated const ID and the CRC-32'd version being proposed

book/src/user/serialization.md
conradoplg: I simply hadn't though of that, good idea. The only catch is that we'd need to change the `ID` in the Ciphersuite to be the contexString instead of the ID (and maybe change its name to `CONTEXT_STRING` too?) - otherwise we'd need to add yet another field to the ciphersuite which doesn't seem great. What do you think?

frost-core/src/lib.rs
conradoplg: Good point. I don't even remember but I guess my thinking was allowing Ciphersuites to override the ID if they wanted to, but I'm not sure if this is actually needed. So I removed it in 8298541 , turning it into a const fn outside the trait.

book/src/user/serialization.md
natalieesk: I think Ciphersuite ID string is fine. Not sure what a better name could be either

conradoplg: I'll improve this later in https://github.com/ZcashFoundation/frost/issues/534

conradoplg: I'll revisit this in https://github.com/ZcashFoundation/frost/issues/534

PR #524, opened 2023-09-05
Protocol state machine implementation for discussion

### LINKED ISSUE #501: high level state machine
currently it requires quite a bit of logic to hook the crypto up into a state machine. we've implemented roast/dkg/recovery helper/recovery state machines and a root level state machine.

The root level state machine api looks something like this:
```
new(peer, members, threshold) -> Self
recover(peer, members, threshold, commitment, public_key_package) -> Self
sign(id, data)
on_message(peer, message)
next_action() -> Option<Action>
```

This is something that is probably useful for many consumers of frost and would probably benefit from some additional eyes. Should I open a preliminary PR next week to discuss if this is something that is in scope for this repo?

dvc94ch:
Builds on the expose-internals PR. Interested to know what you think about doing something like this here. Closes #501.

This is the top level api. P is the network peer identifier and I is the signing session identifier. When a commitment is provided a key share recovery process is started. If no commitment is provided a key generation process is started. For our use case we have shards and tasks in our substrate runtime. A shard member runs the tasks assigned to the shard and then uses this state machine to sign the result of the task. During key generation the commitment is submitted on chain, the proof of knowledge is verified on chain and the group commitment is computed on chain. Once the shard has a group commitment, on_commit is called to complete the other rounds of the key generation.  Once the key generation is complete a public key is returned and the members notify the chain that the shard is ready to process tasks.

```rust
impl Tss<P, I> {
	pub fn new(
	        peer_id: P,
		members: BTreeSet<P>,
		threshold: u16,
		commitment: Option<VerifiableSecretSharingCommitment>,
	) -> Self;
        pub fn on_request(
		&mut self,
		peer_id: P,
		request: TssRequest<I>,
	) -> Result<Option<TssResponse<I>>>;
        pub fn on_response(&mut self, peer_id: P, response: Option<TssResponse<I>>);
        pub fn on_commit(&mut self, commitment: VerifiableSecretSharingCommitment);
        pub fn on_sign(&mut self, id: I, data: Vec<u8>);
        pub fn next_action(&mut self) -> Option<TssAction<I, P>>;
}

#[derive(Clone)]
pub enum TssAction<I, P> {
	Send(Vec<(P, TssRequest<I>)>),
	Commit(VerifiableSecretSharingCommitment, ProofOfKnowledge),
	PublicKey(VerifyingKey),
	Signature(I, [u8; 32], Signature),
}

#[derive(Clone, Deserialize, Serialize)]
pub enum TssRequest<I> {
	Dkg { msg: DkgMessage },
	Rts { msg: RtsRequest },
	Roast { id: I, msg: RoastRequest },
}

#[derive(Clone, Deserialize, Serialize)]
pub enum TssResponse<I> {
	Rts { msg: RtsResponse },
	Roast { id: I, msg: RoastSignerResponse },
}
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,5 +8,6 @@ members = [
     "frost-ristretto255",
     "frost-secp256k1",
     "frost-rerandomized",
-    "gencode"
+    "gencode",
+    "tss",
 ]
--- a/frost-core/src/frost/keys.rs
+++ b/frost-core/src/frost/keys.rs
@@ -50,7 +50,7 @@ pub(crate) fn compute_verifying_share<C: Ciphersuite>(
     group_commitment: &VerifiableSecretSharingCommitment<C>,
     peer: Identifier<C>,
 ) -> VerifyingShare<C> {
-    VerifyingShare::new(evaluate_vss(group_commitment, peer))
+    VerifyingShare(evaluate_vss(group_commitment, peer))
 }
 
 /// Computes the group public key given the group commitment.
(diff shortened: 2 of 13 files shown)


dvc94ch:
@conradoplg any comments? The whole thing is out of date already and we haven't pushed it into production yet, but the basic idea should be there.  For example roast sessions shouldn't be terminated until on_complete is called.

conradoplg:
It's interesting! I need to take a closer look at it and will get back to you.

PR #539, opened 2023-09-15
Rename fields to match types

### LINKED ISSUE #495: Rename fields to match types
We did a bunch of type renamings but forgot to rename fields that used them.

For example, `signer_pubkeys` which is a map of `VerifyingShare`s so it should probably be named `verifying_shares`; `group_public` which is a `VerifyingKey` and should be named `verifying_key` and so on.

Rename all fields to match the types where that makes sense.

This would be a big refactoring so we should do it after all audit issues are merged.

natalieesk:
This closes #495 

1. Updated SecretShare, Package and KeyPackage fields to use signing_share
2. Updated PublicKeyPackage to use verifying_shares
3. Updated KeyPackage to use verifying_key
4. Removed secret() from SecretShare


conradoplg (approved):
This is great, thank you!

Could you please add a CHANGELOG entry for this, listing the old and new names, e.g. "SecretShare::value()" was changed to SecretShare::signing_share()"?

I'm approving this so it will get merged soon, to unblock other work, you can do that in a separate PR if it's OK.

PR #540, opened 2023-09-15
add version field to serialization at the beginning; move ciphersuite after it

### LINKED ISSUE #534: Add version field to serialization at the beginning; move ciphersuite after it
Currently we have a ciphersuite ID at the end.

We should have a version field so that it's possible to upgrade the format if needed (it would be a pain to support two formats but at least there would be the possibility to do that). And it just looks better to have the ciphersuite go after it so that we have a fixed "header" of sorts.

- Add a version field (u8? u32?) at the start, hardcoded with 0 for now
  - Deserialization should check it's zero, so this will need special handling like the ciphersuite ID
- Move ciphersuite ID after it

conradoplg:
This also adds a Header field to the structs as dicussed.

This also changes the Ciphersuite::ID to be the contextString from the spec, per @dconnolly 's suggestion. My arguments in favor of that are:

- It's something concrete from the spec, rather then the old value which does not have a specific name
- It includes a version. If the ciphersuite in the spec changes for whatever reason and contextString changes, then it's basically a new ciphersuite and any serialized FROST structs for that ciphersuite will be incompatible

I _didn't_ change Ciphersuite::ID to be Ciphersuite::CONTEXT_STRING because it doesn't seem to gain us much, but I'm still undecided, feedback is welcome.

Closes #534


book/src/user/serialization.md
```diff
@@ -4,7 +4,8 @@ With the `serialization` feature, which is enabled by default, all structs that
 need to communicated will have `serialize()` and `deserialize()` methods.
 
 The format is basically the `serde` encoding of the structs using the
-[`postcard`](https://docs.rs/postcard/latest/postcard/) crate.
+[`postcard`](https://docs.rs/postcard/latest/postcard/) crate. But since this
+a implementation detail, we describe the format as follows:
```
natalieesk, line 8:
```suggestion
[`postcard`](https://docs.rs/postcard/latest/postcard/) crate. But since this is
an implementation detail, we describe the format as follows:
```

frost-core/src/lib.rs
```diff
@@ -252,7 +254,10 @@ where
 ///
 /// [FROST ciphersuite]: https://www.ietf.org/archive/id/draft-irtf-cfrg-frost-14.html#name-ciphersuites
 pub trait Ciphersuite: Copy + Clone + PartialEq + Debug {
-    /// The ciphersuite ID string
+    /// The ciphersuite ID string. It should be equal to the contextString in
+    /// the spec. For new ciphersuites, this should be a string the identifies
```
natalieesk, line 258:
```suggestion
    /// the spec. For new ciphersuites, this should be a string that identifies
```

natalieesk (changes requested):
Just a few typos but otherwise looks good.

natalieesk (approved):

Issue #525, opened 2023-09-05
Generate randomizer by hashing SigningPackage


conradoplg:
Currently both `frost-rerandomized` and the [ZIP draft](https://github.com/zcash/zips/pull/662) generate randomizers by simply calling `Field::random()` however it would be good to hash information about the ongoing FROST signing into the randomizer generator. This is also part of the security proof of rerandomized FROST. As Chelsea described:

>  I'm assuming participants randomize before performing the second signing round, so aux can include the message, the set of signers, and commitments from all participants from the first signing round

This basically entails hashing the `SigningPackage`. With #511 we can simply use its default serialization.

The final result would look like:

- Generate a random scalar with `Field::random()`
- Add Yet Another Hash Function to the Ciphersuite (i.e. `HRandomizer`) which hashes to a scalar
- Call it by passing the serialized random scalar concatenated with the serialized SigningPackage

This will require changing the `RandomizedParams::new()` to take a `&SigningPackage`.

- Change frost-rerandomized as described
- Update the ZIP to reflect this

conradoplg:
ZIP part done in https://github.com/zcash/zips/pull/662/commits/b441df0745dd0840c5b02bf5dde3955c7640b434

PR #541, opened 2023-09-18
remove serde bounds from Ciphersuites

conradoplg:
When using `derive`, serde adds bounds to (De)Serialize even to the Ciphersuite itself, even though that's not required. That is annoying because it requires us to add the bounds to all ciphersuite structs, but more importantly, is making things much harder in #525 since it needs to use the new default serialization methods which require serde.

This uses `serde(bound = "C: Ciphersuite")` to remove those bounds. See https://stackoverflow.com/questions/76107066/how-to-help-deserializer-understand-that-generics-dont-need-deserialize-bound.

Part of #525 but does not close it.


natalieesk (approved):
Looks good

PR #546, opened 2023-09-21
dependabot: remove hardcoded reviewer and assignee

conradoplg:
I figured we can just omit it and assign it ourselves per availability


natalieesk (approved):
Looks good

PR #547, opened 2023-09-22
Convert all HashMaps to BTreeMaps

### LINKED ISSUE #476: Replace all HashMaps with BTreeMaps
We use both, but we use BTreeMap in some places since it iterates in order which is useful for FROST. For consistency we might want to change all HashMaps to BTreeMaps.

natalieesk:
This closes #476


conradoplg (approved):
Looks good, thank you!

PR #484, opened 2023-08-17
fix DKG challenge hashing order to match paper

conradoplg:
Closes https://github.com/ZcashFoundation/frost/issues/462

PR #551, opened 2023-09-26
add computing PublicKeyPackage from commitments

conradoplg:
Allow computing PublicKeyPackage from commitments.

It also refactors a lot of related code which makes everything more organized.

Expanded from https://github.com/ZcashFoundation/frost/pull/502/, I made a mess with the history after syncing with main and didn't want to force push in the contributor's PR.


frost-core/src/frost/keys.rs
```diff
@@ -29,6 +29,35 @@ use super::compute_lagrange_coefficient;
...
+/// run into a single group commitment.
+#[cfg_attr(feature = "internals", visibility::make(pub))]
+pub(crate) fn sum_commitments<C: Ciphersuite>(
+    commitments: &[&VerifiableSecretSharingCommitment<C>],
+) -> Result<VerifiableSecretSharingCommitment<C>, Error<C>> {
+    let mut group_commitment = vec![
+        CoefficientCommitment(<C::Group>::identity());
+        commitments
+            .get(0)
+            .ok_or(Error::IncorrectCommitment)?
```
natalieesk, line 42: I feel like a more appropriate error here would be an InvalidCommitment as IncorrectCommitment is more about a commitment as it relates to a participant. Just a thought but not crucial.

conradoplg: I think having both Invalid and Incorrect would be confusing, it doesn't seem clear what would be their difference. I went for reusing IncorrectNumberOfCommitments in 7965e24 which seems to be the actual issue here (either `commitments` is empty, or the one commitment vector has a different number of commitments). What do you think?

conradoplg (commented):
Addressed comment

conradoplg:
Last Rust also introduced some clippy issues so I went ahead and fixed them here.

frost-core/src/frost/keys/dkg.rs
```diff
@@ -331,6 +328,56 @@ where
...
+    // Round 1, Step 2
+    //
+    // > Every P_i computes a proof of knowledge to the corresponding secret
+    // > a_{i0} by calculating σ_i = (R_i, μ_i), such that k ← Z_q, R_i = g^k,
+    // > c_i = H(i, Φ, g^{a_{i0}} , R_i), μ_i = k + a_{i0} · c_i, with Φ being
+    // > a context string to prevent replay attacks.
+    let k = <<C::Group as Group>::Field>::random(&mut rng);
+    let R_i = <C::Group>::generator() * k;
+    let c_i =
+        challenge::<C>(identifier, &R_i, &commitment.first()?.0).ok_or(Error::DKGNotSupported)?;
```
natalieesk, line 349: Are the verifying_key and commitment the wrong way round here? I think it's meant to be 
```let c_i =challenge::<C>(identifier, &commitment.first()?.0, &R_i).ok_or(Error::DKGNotSupported)?;```

File: frost-core/src/frost/keys/dkg.rs [hunk 3]
```diff
@@ -331,6 +328,56 @@ where
...
+    // Round 1, Step 5
+    //
+    // > Upon receiving C⃗_ℓ, σ_ℓ from participants 1 ≤ ℓ ≤ n, ℓ ≠ i, participant
+    // > P_i verifies σ_ℓ = (R_ℓ, μ_ℓ), aborting on failure, by checking
+    // > R_ℓ ? ≟ g^{μ_ℓ} · φ^{-c_ℓ}_{ℓ0}, where c_ℓ = H(ℓ, Φ, φ_{ℓ0}, R_ℓ).
+    let ell = identifier;
+    let R_ell = proof_of_knowledge.R;
+    let mu_ell = proof_of_knowledge.z;
+    let phi_ell0 = commitment.first()?.0;
+    let c_ell = challenge::<C>(ell, &R_ell, &phi_ell0).ok_or(Error::DKGNotSupported)?;
```
natalieesk, line 374: Same here. I think R and phi should be the other way round.

frost-core/src/frost/keys/dkg.rs
conradoplg: Well spotted, thank you! I fixed this in 28c582d. To be sure we get it right even if refactor things, I changed the `challenge` function to take a VerifyingKey instead of an Element. Then I also changed the `first()` method to `verifying_key()` which also has the advantage of being more descriptive.

I also went ahead and did the same thing for the regular FROST `challenge()` function in b1c3989, looks much nicer.

Lesson learned, I'll try to avoid writing functions with multiple arguments of the same type in the future.

frost-core/src/frost/keys/dkg.rs
conradoplg: Same as above

frost-core/src/frost/keys/dkg.rs
conradoplg: Also here is the PR were we fixed the order, to double check we're hashing in the correct order https://github.com/ZcashFoundation/frost/pull/484/files

natalieesk (approved):
Looks good

PR #554, opened 2023-10-02
update README, move dev info to book

### LINKED ISSUE #510: Update READMEs for 1.0.0 release
The READMEs have some information that is or will become outdated, such as the code not being audited.

We should update them all and:

- Add docs, CI, etc "buttons" like most Rust crates have
- Point to the book for documentation

conradoplg:
Closes #510


natalieesk (approved):
Looks good

PR #555, opened 2023-10-02
initial cargo vet support

conradoplg:
I wanted to give `cargo vet` a try. The first step is to run `cargo vet init` which basically marks all current dependencies as trusted. (That information is written to the files that this PR adds) Then as they get updated we can use `cargo vet` to help audit them.

This has no concrete impact in our workflow until CI is not changed to run `cargo vet`, which we can do later if we decide to commit to using `cargo vet`.

I also marked some orgs as trusted for audits - Google, Mozilla and ECC

### DIFF
--- a/supply-chain/audits.toml
+++ b/supply-chain/audits.toml
@@ -1,0 +1,4 @@
+
+# cargo-vet audits file
+
+[audits]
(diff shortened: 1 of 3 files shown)


conradoplg:
I'm closing this for now, we need to figure out how to incorporate it into our workflow

PR #564, opened 2023-10-18
Support optional cheater detection

### LINKED ISSUE #355: Support optional cheater detection
Cheater detection might not be wanted. In that case, the aggregator doesn't need to store all the participants VerifyingShares. We can support that by changing `aggregate()` to not panic if it doesn't find a VerifyingShare (or maybe better, by simply skipping the cheat detection loop if the list of VerifyingShares is empty)

natalieesk:
Adds cheater detection feature which gives the option to use cheater detection in the aggregate function or not.

This closes #355

### DIFF
--- a/frost-core/Cargo.toml
+++ b/frost-core/Cargo.toml
@@ -63,6 +63,8 @@ serde = ["dep:serde", "dep:serdect"]
 serialization = ["serde", "dep:postcard"]
 # Exposes ciphersuite-generic tests for other crates to use
 test-impl = ["proptest", "serde_json", "criterion"]
+# Enables cheater detection
+cheater-detection = []
 
 [lib]
 bench = false
--- a/frost-core/Cargo.toml
+++ b/frost-core/Cargo.toml
@@ -50,7 +50,7 @@ rand_chacha = "0.3"
 serde_json = "1.0"
 
 [features]
-default = ["serialization"]
+default = ["serialization", "cheater-detection"]
 #! ## Features
 ## Expose internal types, which do not have SemVer guarantees. This is an advanced
 ## feature which can be useful if you need to build a modified version of FROST.
@@ -63,7 +63,7 @@ serde = ["dep:serde", "dep:serdect"]
 serialization = ["serde", "dep:postcard"]
 # Exposes ciphersuite-generic tests for other crates to use
 test-impl = ["proptest", "serde_json", "criterion"]
-# Enables cheater detection
+# Enable cheater detection
 cheater-detection = []
 
 [lib]
--- a/frost-core/src/frost.rs
+++ b/frost-core/src/frost.rs
@@ -406,6 +406,8 @@ where
 /// signature, if the coordinator themselves is a signer and misbehaves, they
 /// can avoid that step. However, at worst, this results in a denial of
 /// service attack due to publishing an invalid signature.
+
+#[cfg(feature = "cheater-detection")]
 pub fn aggregate<C>(
     signing_package: &SigningPackage<C>,
     signature_shares: &BTreeMap<Identifier<C>, round2::SignatureShare<C>>,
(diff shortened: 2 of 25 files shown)


conradoplg (commented):
I assumed that this would be a runtime thing, not compile time. Sorry for not clearly communicating that. (To do at runtime it could check if the verifying shares map is empty, and disable it if it is)

However, this is a valid way to do it. I'm not sure about it though. One advantage of being compile-time is that it's harder to accidentally disable it (assuming we enable it by default, which this PR doesn't do yet, but it's a simple change).

On the other hand we could find a way to do it at runtime while making it harder to disable by accident. For example, with a new `aggregate_without_cheater_detection()` function (that would call an e.g. refactored out private `aggregate_inner()` function that is also called by `aggregate()`).

But I'm not sure if we _need_ to do it at runtime.

What do you think?

conradoplg:
> is it PR title vague or I'm reading it wrong? pr gates existing cheater detection under a feature gate but titled as like cheater detection functionality?

Yes, if you look at the matching issue, it's about making cheater detection optional. I've updated the PR title to match it.

frost-core/src/frost.rs
```diff
@@ -506,3 +508,78 @@ where
...
+/// coordinator can be one of the participants or a semi-trusted third party
+/// (who is trusted to not perform denial of service attacks, but does not learn
+/// any secret information). Note that because the coordinator is trusted to
+/// report misbehaving parties in order to avoid publishing an invalid
+/// signature, if the coordinator themselves is a signer and misbehaves, they
+/// can avoid that step. However, at worst, this results in a denial of
+/// service attack due to publishing an invalid signature.
+
+#[cfg(not(feature = "cheater-detection"))]
+pub fn aggregate<C>(
```
conradoplg: I'd like to avoid duplicated code here. The `#[cfg()]` attribute works for statements too, so we can deactivate parts of the function. I tried this approach and it seems to work, I comitted it in this branch in 325cb03. Can you take a look and review it in case you like the idea?

frost-ed25519/Cargo.toml
```diff
@@ -52,6 +52,8 @@ serialization = ["serde", "frost-core/serialization"]
 ## can use `serde` to serialize structs with any encoder that supports
 ## `serde` (e.g. JSON with `serde_json`).
 serde = ["frost-core/serde"]
+## Enable cheater detection
+cheater-detection = ["frost-core/cheater-detection"]
```
conradoplg, line 56: We will also need to enable it by default in the ciphersuite crates

frost-ed25519/tests/integration_tests.rs
```diff
@@ -8,6 +8,15 @@ fn check_zero_key_fails() {
...
 
+#[cfg(feature = "cheater-detection")]
+#[test]
+fn check_sign_with_dkg() {
+    let rng = thread_rng();
+
+    frost_core::tests::ciphersuite_generic::check_sign_with_dkg::<Ed25519Sha512, _>(rng);
+}
+
+#[cfg(not(feature = "cheater-detection"))]
```
conradoplg, line 19: I don't get this, both options seems to be the same? Maybe leftover code?

frost-ed25519/tests/integration_tests.rs
```diff
@@ -64,6 +73,15 @@ fn check_rts() {
...
 
+#[cfg(feature = "cheater-detection")]
+#[test]
+fn check_sign_with_dealer() {
+    let rng = thread_rng();
+
+    frost_core::tests::ciphersuite_generic::check_sign_with_dealer::<Ed25519Sha512, _>(rng);
+}
+
+#[cfg(not(feature = "cheater-detection"))]
```
conradoplg, line 84: Same here

frost-ed25519/tests/integration_tests.rs
```diff
@@ -205,6 +223,18 @@ fn check_identifier_generation() -> Result<(), Error> {
...
+fn check_sign_with_dealer_and_identifiers() {
+    let rng = thread_rng();
+
+    frost_core::tests::ciphersuite_generic::check_sign_with_dealer_and_identifiers::<
+        Ed25519Sha512,
+        _,
+    >(rng);
+}
+
+#[cfg(not(feature = "cheater-detection"))]
```
conradoplg, line 237: Same here

conradoplg (changes requested):
I comitted a suggestion and there are some comments.

The PR will unbreak after #551 merges

frost-core/Cargo.toml
```diff
@@ -63,6 +63,8 @@ serde = ["dep:serde", "dep:serdect"]
 serialization = ["serde", "dep:postcard"]
 # Exposes ciphersuite-generic tests for other crates to use
 test-impl = ["proptest", "serde_json", "criterion"]
+# Enables cheater detection
+cheater-detection = []
```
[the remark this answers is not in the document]
natalieesk: Done in [541ebf3](https://github.com/ZcashFoundation/frost/pull/564/commits/541ebf338787812c5dc935cd6b8658b03ca7eff3)

frost-core/src/frost.rs
natalieesk: Yes, much better. I did that for the tests but for some reason not for the code. Refactoring oversight!

frost-ed25519/tests/integration_tests.rs
natalieesk: They all call check_sign() in the tests which is where the feature is toggled. https://github.com/ZcashFoundation/frost/blob/cheater_detection_355/frost-core/src/tests/ciphersuite_generic.rs#L251. So these tests are one with the feature enabled and one without. Is there a nicer way of writing this?

frost-ed25519/Cargo.toml
natalieesk: Done in [4d26417](https://github.com/ZcashFoundation/frost/pull/564/commits/4d26417d1d039f2fd9ba38b3b7f149cbd8a7e297)

frost-ed25519/tests/integration_tests.rs
natalieesk: All these functions call check_sign() which has the feature toggled

conradoplg: Ah, I see what was your intention. This will not quite work that way, the `cfg()` attribute will just enable some piece of code depending on whether the feature was enabled or not, it won't actually enable or disable it.

Here both functions are the same, so there is no need to gate them under `cfg()`. Whether it will call the `check_sign()` with the feature enabled or not will only depend on whether the feature is enabled in the first place.

So you can remove the `cfg()` gate there and leave it like it was before.

In order to actually test the case where the feature is disabled, we'll need to call `cargo test --release --no-default-features` in CI (by adding a new entry in `main.yml`). You can do that in this PR or in a separate one, up to you.

(Ideally we should test all possible combinations of features. I think there is a cargo plugin that helps with that. But that's probably overkill in our case)

natalieesk: Ah ok. Thank you. I'll update the code

conradoplg (approved):
Looks good, thanks!

PR #568, opened 2023-10-30
remove frost module

### LINKED ISSUE #490: Remove frost module
In frost-core we currently have

- crate: non-FROST types for signing and verifying
- crate::frost: FROST types
- crate::frost::{keys, round1, round2}: the rest

However ciphersuite crates have

- crate: non-FROST types for signing and verifying, and FROST types
- crate::{keys, round1, round2}: the rest

Change frost-core to follow the ciphersuite crates, merging `crate::frost` into `crate::` and removing `crate::frost` and move child modules up accordingly.

This is a significant (though simple) refactoring so we should do this after all open PRs are merged

conradoplg:
Closes #490 

This:

- Merges `frost_core/src/frost.rs` into `frost_core/src/lib.rs`
- Moves everything in `frost_core/src/frost/*` into `frost_core/src/`
- I also updated the README a bit

I think `lib.rs` is now too big. I want to split it into files (while keeping the API the same) but I'll do that in a follow up PR.


natalieesk (approved):
Looks good

PR #569, opened 2023-10-31
frost-core: split part of lib.rs into traits.rs and serialization.rs

conradoplg:
As mentioned in #568 , `lib.rs` got too big so I split parts of it into `traits.rs` and `serialization.rs`.

While doing that I noticed that some types were pub but they didn't need to be, so I hid them.


natalieesk (approved):
Looks good

Issue #519, opened 2023-09-04
Allow signers to change group threshold by updating shares


conduition:
There is [this interesting protocol](https://uwspace.uwaterloo.ca/bitstream/handle/10012/6858/nojoumian_mehrdad.pdf?sequence=1#subsection.4.3.1) which allows a group of at least $t$ shamir shareholders (or FROST signers) to compute updated shares which correspond to a new threshold $t'$. 

It requires that only $t$ signers be online, but all $n$ signers must at least be able to receive messages asynchronously while offline, to update their shares to have threshold $t'$. Of course it wouldn't stop a malicious subset of shareholders from holding onto their old shares which are valid for the original threshold $t$. 

Is this of interest? Would love to work on this. 

### Here's a short description, omitting verifiable commitments.

- Let $f(x)$ be the joint secret polynomial. In the FROST paper this is referred to as $F_1(x)$.  
- Let $P_i$ be the participant at index $i$. Each $P_i$ knows a secret share $z_i = f(i)$.
- Let $S$ be a set of $t$ or more share indexes, held by online participants.
- Let $\gamma_i^X$ be the lagrange coefficients for a set of inputs $X$, used to evaluate a polynomial at $0$:

$$\gamma_i^X = \prod_{\substack{j \in X \\\\ j \ne i}} \frac{0 - j}{i - j}$$

1. Each online participant $P_i : i \in S$ samples $t' - 1$ random coefficients $\\{a_1, a_2, ..., a_{t' - 1}\\}$. This defines the polynomial

$$g_i(x) = z_i + \sum_{j=1}^{t'-1} a_j x^j$$

In other words, each $g_i(x)$ is a random degree $t' - 1$ polynomial with $z_i = f(i)$ as the constant term. 

2. Online participant $P_i$ sends $g_i(j)$ to every participant $P_j$ for every $j \in \\{1...n\\}$. It is important that the offline participants can receive these messages asynchronously, as they will need to update their shares before their next interaction with any shareholders in $S$.
3. $P_i$ receives all $\\{g_j(i)\\}\_{j \in S}$.

4. $P_i$ computes his new share $z_i'$ as:

$$z_i' = \sum_{j \in S} \gamma_j^S g_j(i)$$

...and erases his old share $z_i$.

5. At recovery time, the shareholders now need a subset $R$ of at least $t'$ shares to recover $f(0)$.

$$
\begin{align}
f(0) &= \sum_{i \in R} \gamma_i^R z_i' \\\\
[...]

Issue #245, opened 2023-02-15
Feature: remove signers or refresh shares with trusted dealer


natalieesk:
## Motivation
There may be situations where you may want to _remove_ signers. 

## Design
To do this you would need to perform a resharing protocol and leave out the relevant participant.

This is done by basically redoing the trusted dealer generation with a "zero" secret, and adding the new shares to the old ones, while keeping only the participants that are wanted

## References
<add here>

chelseakomlo:
I changed this issue to "remove signers or refresh shares" since the technique is the same.

chelseakomlo:
Completed

mpguerra:
@natalieesk can you please split this one up and we turn it into a tracking issue instead

PR #570, opened 2023-11-06
add verifiable key resharing module

conduition:
This PR implements Wang's Verifiable Secret Redistribution (VSR) Scheme to allow resharing of FROST shares, with protection against active adversaries. See https://github.com/ZcashFoundation/frost/issues/519 for research and justification. Also see https://github.com/ZcashFoundation/frost/issues/245 for related work.

This enables threshold modification, signer identifier changes, share revocation, and other cool downstream use-cases. 

This PR is not complete; i wanted to check in and get feedback before proceeding further. 

Remaining work:

- [ ] More thorough integration tests
- [ ] Implement the `resharing` module in all `frost-*` ciphersuite crates.
- [ ] Improve documentation.
- [x] Add docs to the book.
- [ ] Other things I'm missing?

### DIFF
--- a/book/book.toml
+++ b/book/book.toml
@@ -9,7 +9,7 @@ title = "The ZF FROST Book"
 
 [preprocessor.admonish]
 command = "mdbook-admonish"
-assets_version = "2.0.1" # do not edit: managed by `mdbook-admonish install`
+assets_version = "3.0.0" # do not edit: managed by `mdbook-admonish install`
 
 [output]
 
--- a/book/mdbook-admonish.css
+++ b/book/mdbook-admonish.css
@@ -1,31 +1,18 @@
 @charset "UTF-8";
 :root {
-  --md-admonition-icon--note:
-    url("data:image/svg+xml;charset=utf-8,<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 24 24'><path d='M20.71 7.04c.39-.39.39-1.04 0-1.41l-2.34-2.34c-.37-.39-1.02-.39-1.41 0l-1.84 1.83 3.75 3.75M3 17.25V21h3.75L17.81 9.93l-3.75-3.75L3 17.25z'/></svg>");
-  --md-admonition-icon--abstract:
-    url("data:image/svg+xml;charset=utf-8,<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 24 24'><path d='M17 9H7V7h10m0 6H7v-2h10m-3 6H7v-2h7M12 3a1 1 0 0 1 1 1 1 1 0 0 1-1 1 1 1 0 0 1-1-1 1 1 0 0 1 1-1m7 0h-4.18C14.4 1.84 13.3 1 12 1c-1.3 0-2.4.84-2.82 2H5a2 2 0 0 0-2 2v14a2 2 0 0 0 2 2h14a2 2 0 0 0 2-2V5a2 2 0 0 0-2-2z'/></svg>");
-  --md-admonition-icon--info:
-    url("data:image/svg+xml;charset=utf-8,<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 24 24'><path d='M13 9h-2V7h2m0 10h-2v-6h2m-1-9A10 10 0 0 0 2 12a10 10 0 0 0 10 10 10 10 0 0 0 10-10A10 10 0 0 0 12 2z'/></svg>");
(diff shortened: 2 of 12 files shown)


conduition:
Thanks for the quick attention @chelseakomlo =) 

-----------

> As a first step, it would be helpful to write a high-level design document to give intuition for the change and how it fits into the user experience/existing code. 

I certainly can. I'll follow the example of the [`dkg.md` document](https://github.com/ZcashFoundation/frost/blob/main/book/src/tutorial/dkg.md).

---------

> I looked over your [notes](https://github.com/ZcashFoundation/frost/issues/519#issuecomment-1722151933) in https://github.com/ZcashFoundation/frost/issues/519 and it lists some [drawbacks](https://www.sciencedirect.com/science/article/abs/pii/S2214212622001491) of the Wang scheme, one which is the lack of public verifiability.

What the author means by 'public verifiability' is a property where **external verifiers** (who are not part of the signing group) can verify that all new shares are distributed accurately and consistently. [This paper](https://link.springer.com/content/pdf/10.1007/BFb0054115.pdf) gives a short and sweet definition of a PVSS in the abstract:


> A publicly verifiable secret sharing (PVSS) scheme, named by Stadler in [Sta96], is a special VSS scheme in which anyone, not only the shareholders, can verify that the secret shares are correctly distributed.

For example, customers of a cryptocurrency exchange may wish to verify that shares of the company's cold-storage signing key are distributed to the appropriate executives. 

FROST does not explicitly support public verifiability (as far as i know) - although it would be trivial to extend it include such features. FROST assumes that only the signers care about whether the signing shares are correctly distributed, while external verifiers outside the FROST signing group are assumed to only care about signature validity against the group's joint pubkey.

Wang's VSR scheme is indeed verifiable in the sense that peers who receive new shares from the VSR protocol - including those who are not already part of the signing group - can verify their new shares will be correct, and accurately assign blame to misbehaving resharer(s). This all assumes the resharing polynomial commitment is sent over a broadcast channel. If commitments are not consistent, then blame is harder to assign accurately.

The group is thus protected against denial-of-service attacks by any $t - 1$ or fewer signers of the old group (before resharing). The group can be DoS'd if $n' - t' + 1$ members of the new group misbehave, but that's exactly the same as normal FROST without resharing, so nothing new there.  



---------

> Also, this link (cited in the code) doesn't open for me- https://www.semanticscholar.org/paper/Verifiable-Secret-Redistribution\-for-Threshold-Wong-Wang/48d248779002b0015bdb99841a43395b526d5f8e

That link should render in docs.rs as https://www.semanticscholar.org/paper/Verifiable-Secret-Redistribution-for-Threshold-Wong-Wang/48d248779002b0015bdb99841a43395b526d5f8e - i think you may have copied the backslash needed to escape the line-break.


I hope this helps, i'll get started on the design document once I have some time

conduition:
[Added a first draft of a document dedicated to resharing.](https://github.com/ZcashFoundation/frost/pull/570/commits/621f79e59e42e255c34a90daffe9968b55f7c151) This doesn't go into detailed code examples as I suspect we may want to modify the API for simplicity/clarity, or to rhyme with the DKG API.

conduition:
Would like some indication I'm going in the right direction here before I press any more effort onto this.

chelseakomlo:
Hi @conduition - sorry for the slow reply, it took us some time to review these and discuss as a team.

While this approach indeed looks promising, the current blocker is that while the resharing algorithm seems to be well-understood for a single-dealer and has a security proof, extending this functionality into a multi-party setting is not necessarily straightforward. As @conradoplg  mentioned, we will need to consider malicious parties, which result in a multi-round DKG protocol, but with additional properties. While I suspect this re-sharing algorithm would slot into existing DKG frameworks [1], someone will need to write this protocol in precise detail and give at minimum a proof sketch that the resulting protocol is indeed a secure DKG, but with the additional claimed properties of adding/removing players and changing the threshold. 

I do think this work is indeed possible, but unfortunately our team currently doesn't have the bandwidth to lead this work at the moment, or to collaborate/provide reviews in a sufficient level of detail at this time. But please feel free to continue development as an external project, it seems useful and something that would be valuable in the future.

Thanks- again, this work looks very promising, sorry we can't focus on it more! And feel free to point to writeups/proofs that might exist already in the literature, if you end up finding something in the future.

[1] https://eprint.iacr.org/2023/292

conduition:
That's OK @chelseakomlo, I understand. Thanks for the transparency and honesty :+1:

PR #573, opened 2023-11-09
re-export serde in ciphersuite crates

conradoplg:
Small cleanup. While working on the demo noticed that we missed this. (Reexporting makes it easier for callers to use the same serde library that our crates use, without having to import it again and figuring out which version to import)


natalieesk (approved):
Looks good

PR #580, opened 2023-11-16
create new() methods for VerifirableSecretSharingCommitment and CoefficientCommitment

conradoplg:
These are needed in `reddsa` to support even Y public keys


natalieesk (approved):
Looks good

PR #584, opened 2023-11-20
Add frost-secp256k1-tr crate (BIP340/BIP341)

zebra-lucky:
- add frost-secp256k1-tr crate with support of BIP0340, BIP0341


conradoplg:
Thank you, this is great. I'll take a closer look soon and may make some adjustments.

conradoplg:
After a closer look, here are some things that we'd prefer to change before merging this PR:

- The current approach changes the key (to ensure a even Y in the public key) during the signing operation. This makes sense if you follow BIP-340 directly. However, with FROST, I think it would be better and simpler to change the key **during share generation**, i.e. if the public key as an odd Y, negate all the shares, verifying shares, etc. This would make the PR much less invasive since the signing code would not change. You can take a [look at this PR](https://github.com/ZcashFoundation/reddsa/pull/94) from the reddsa crate which does basically the same thing
- The PR also tweaks the key per BIP-341 with an empty merkle root. As I understand this wouldn't work in practice since the merkle root needs to be specified. I think this can be accomplished separately, using the `frost-rerandomized` crate - I think that the Taproot tweak is basically equal to rerandomization, but with a specific tweak rather than a random one.

I realize these might be big changes. Let me know if this makes sense and if you are able to work on this. Otherwise we can eventually incorporate these changes to the PR, but might not be able to get that done shortly.

Also - what did you use to generate the test vectors? Do you have a separate reference implementation for this? I'd like to take a look if possible.

zebra-lucky:
> I realize these might be big changes. Let me know if this makes sense and if you are able to work on this. Otherwise we can eventually incorporate these changes to the PR, but might not be able to get that done shortly.

I'll try to add changes to generation process to use only even VerifyingKey

On the "emtpy merkle root" -- need to do more investigations, but current code is working. As a proof I'll add link on the repo with testing code to sign tx (in the next post).

> Also - what did you use to generate the test vectors? Do you have a separate reference implementation for this? I'd like to take a look if possible.

On testing vectors -- I just added the values from the working code so the tests wouldn't fail.

zebra-lucky:
Some proofs that the current code works

Testing code on the Testnet `sign-tx-frost` https://github.com/zebra-lucky/sign-tx-frost

To use there is a need to fix paths in the `Cargo.toml`
```
frost-core = { version = "1.0.0-rc.0", features = ["serde"], path = "../frost.copy/frost-core" }
...
frost-secp256k1-tr = { version = "1.0.0-rc.0", features = ["serde"], path = "../frost.copy/frost-secp256k1-tr"  }
```

- `generate` command outputs to stdout and this can be saved in `testdata.json`
- `address` command show address on generated `VerifyinKey` in the `testdata.json`
- `sendtoaddress` allow sign and get hex of output transaction with arguments in the form:

```sendtoaddress tb1pqkgsz274gjnkdxp7v9rpzwqtqtjacjp5t2mz2vaqu2r6qln8qsesve6uez 1000 02000000000101bf2d2d426...```

Where last part is a hex of transaction which makes output on generated address.

Some dirty but working code.

Latest transactions on the Testnet:
https://live.blockcypher.com/btc-testnet/address/tb1pqkgsz274gjnkdxp7v9rpzwqtqtjacjp5t2mz2vaqu2r6qln8qsesve6uez/

File: frost-core/src/round2.rs [hunk 1]
```diff
@@ -90,9 +90,23 @@ where
...
+        let mut commitment_share = group_commitment_share.0;
+        let mut vsh = verifying_share.0;
+        if <C>::is_need_tweaking() {
+            commitment_share = <C>::tweaked_group_commitment_share(
+                &group_commitment_share.0,
+                &group_commitment.0
+            );
+            vsh = <C>::tweaked_verifying_share(
+                &verifying_share.0,
+                &verifying_key.element
```
mimoo, line 105: these "tweaks" have a different meaning as the taproot tweaks right? I'm confused :D

mimoo:
Just FYI we're testing this library and getting flaky results. I'm investigating on our side to understand if it's us or if there's some issue here

frost-core/src/verifying_key.rs
```diff
@@ -63,9 +68,15 @@ where
...
+        let mut R = signature.R;
+        let mut vk = self.element;
+        if <C>::is_need_tweaking() {
+            R = <C>::tweaked_R(&signature.R);
+            vk = <C>::tweaked_public_key(&self.element);
+        }
         let zB = C::Group::generator() * signature.z;
-        let cA = self.element * challenge.0;
-        let check = (zB - cA - signature.R) * C::Group::cofactor();
+        let cA = vk * challenge.0;
```
mimoo, line 78: OK so I think there's an issue here, vk might be the wrong tweaked public key  as `<C>::tweaked_public_key(&self.element);` on line 75 computes a tweaked public key point on the fly. But the tweaked public key point that Bitcoin might use is the one that is taken from the x coordinate only!

frost-secp256k1-tr/src/lib.rs
```diff
@@ -0,0 +1,628 @@
...
+/// Create a BIP341 compliant tweaked public key
+pub fn tweaked_public_key(
+    public_key: &<<Secp256K1Sha256 as Ciphersuite>::Group as Group>::Element,
+    merkle_root: &[u8],
+) -> <<Secp256K1Sha256 as Ciphersuite>::Group as Group>::Element {
+    let mut pk = public_key.clone();
+    if public_key.to_affine().y_is_odd().into() {
+        pk = -pk;
+    }
+    ProjectivePoint::GENERATOR * tweak(&pk, merkle_root) + pk
```
mimoo, line 229: notice that this might be the wrong tweaked public key, as it returns a `(x, y)` but in bitcoin only the x coordinate is given and so the public key used to verify the signature might be `(x, -y)`. I think you need to add

```rust
let tweaked_pubkey = ProjectivePoint::GENERATOR * tweak(&pk, merkle_root) + pk;
AffinePoint::decompact(&tweaked_pubkey.to_affine().x()).unwrap().into()
```

frost-secp256k1-tr/src/lib.rs
```diff
@@ -0,0 +1,628 @@
...
+
+    /// aggregate tweak z
+    fn aggregate_tweak_z(
+        z: <<Self::Group as Group>::Field as Field>::Scalar,
+        challenge: &Challenge<S>,
+        verifying_key: &Element<S>,
+    ) -> <<Self::Group as Group>::Field as Field>::Scalar
+    {
+        let t = tweak(&verifying_key, &[]);
+        z + t * challenge.clone().to_scalar()
```
mimoo, line 331: I think at this point there's two different cases:

* `z + t * c`
* `z - t * c`

mimoo: so maybe this should be the code IIUC:

```rust
let tc = t * challenge.clone().to_scalar();
if verifying_key.y_is_odd() { z - tc } else { z + tc }
```

frost-secp256k1-tr/src/lib.rs
```diff
@@ -0,0 +1,628 @@
...
+    {
+        let mut sn = signer_nonces.clone();
+        if group_commitment.y_is_odd() {
+            sn.negate_nonces();
+        }
+
+        let mut kp = key_package.clone();
+        if key_package.verifying_key().y_is_odd() {
+            kp.negate_signing_share();
+        }
```
mimoo, line 352: IIUC this should also use the y_is_odd of the tweaked public key!

```rust
let tweaked_pubkey = tweaked_public_key(key_package.verifying_key().element(), &[]);
        if tweaked_pubkey.to_affine().y_is_odd().into() {
            kp.negate_signing_share();
        }
```

mimoo: or this might be it:

```rust
let public_key = key_package.verifying_key();
let pubkey_is_odd = public_key.y_is_odd();
let tweaked_pubkey_is_odd = tweaked_public_key(publickey.element(), &[]).is_odd().into();
if pubkey_is_odd != tweaked_pubkey_is_odd {
  kp.negate_signing_share();
}
```

mimoo:
I fixed all the comments I pointed out in https://github.com/zebra-lucky/frost/pull/1 and everything now works on our end!

frost-core/src/round2.rs
zebra-lucky: Hi! I'm sorry for a pause for reply.

Yes, it's not taproot, maybe need more correct names to `fn`.

zebra-lucky:
> Nevermind, I wrote my understanding here: https://hackmd.io/li6ByhmfQUucZOqdboF1-g?view

Thanks!, i'll read it to realize.

zebra-lucky:
I am looking at the changes proposed in https://github.com/zebra-lucky/frost/pull/1

zebra-lucky:
Merged fix on tweaked public key processing from @mimoo.
https://github.com/zebra-lucky/frost/pull/1

I'll add DKG/trusted dealer code which generates only
even `VerifyingKey` in a few hours.

mimoo:
> I'll add DKG/trusted dealer code which generates only even VerifyingKey

actually that might be overkill since all the checks must happen anyway no?

zebra-lucky:
> actually that might be overkill since all the checks must happen anyway no?

yep... thinking on that...

currently stuck on batch test... signing with `SiginingKey`...

zebra-lucky:
Added:
- Some more fixes on tweaked public key use after https://github.com/zebra-lucky/frost/pull/1 merge
- Renamed fn with more consistent names
- add DKG vector test for frost-secp256k1-tr (as in recent commitments for other curves)

There is a question about only even `VerifyingKey` generation with DKG/trusted dealer.
This code is done, but tweaked public key is independently can be even or odd.
Try to analyse tomorrow, but seems no overall code simplification will be possible.
But maybe some refactoring on current code should be done.

Not about merkle root (as in [BIP341](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki) code example):
```
def taproot_sign_key(
    script_tree,
    internal_seckey,
    hash_type,
    bip340_aux_rand
):
    if script_tree is None:
        h = bytes()
...
```

For key path spending `script_tree` is absent, so empty bytes value set as `h`

zebra-lucky:
p.s.: rebased PR branch on current main branch

frost-secp256k1-tr/src/lib.rs
```diff
@@ -0,0 +1,657 @@
...
+        Some(hash_to_scalar(
+            (CONTEXT_STRING.to_owned() + "id").as_bytes(),
+            m,
+        ))
+    }
+
+    /// Generates the challenge as is required for Schnorr signatures.
+    fn challenge(R: &Element<S>, verifying_key: &VerifyingKey, msg: &[u8]) -> Challenge<S> {
+        let mut preimage = vec![];
+        let tweaked_public_key = tweaked_public_key(&verifying_key.to_element(), &[]);
```
0xBEEFCAF3, line 320: Should the merkel root be empty here?
i.e does this code assume that we only spend from the default key path? No a tapscript path?

zebra-lucky: Possibly my understanding is wrong, but in BIP341, in
code example for `taproot_sign_key` I've seen next:
```
    if script_tree is None:
        h = bytes()  # empty bytes for merkle root
```

In the case when script path is not used (used only key path),
seems empty bytes for merkle root is correct.

mimoo: yeah this code only works with empty commitment, and it works fine for us (https://github.com/sigma0-xyz/zkbitcoin). I think it's a good first step, but it'd be nice to expose that feature later

zebra-lucky: Thank you. I'll look to code in zkbitcoin in a few days.

0xBEEFCAF3: >In the case when script path is not used (used only key path),

Ah yes you're correct. In the case of a script path spend (non-frost spend) the spender can tweak the verifying key outside the context of this library.

Polybius93:
Subject: Error Broadcasting PSBT with FROST Aggregated Signature: Invalid Schnorr Signature

Hello,

I'm encountering an issue when broadcasting a PSBT transaction signed using FROST for aggregated Schnorr signatures. While individual Schnorr signatures work correctly, the aggregated signature fails with the following error:

```
"sendrawtransaction RPC error: {"code":-26,"message":"non-mandatory-script-verify-flag (Invalid Schnorr signature)"}
```

**Process Overview:**

Individual Signing (Works): We generate a hash from the PSBT's unsigned transaction and sign it using a keypair. This process completes without errors.

**Hash generation:**
```rust
let hash = SighashCache::new(&psbt.unsigned_tx.clone())
            .taproot_script_spend_signature_hash(
                0,
                &Prevouts::All(&[tx_out.clone()]),
                ScriptPath::with_defaults(&multisig_script),
                SchnorrSighashType::Default,
            )
            .unwrap_throw();
```
**Schnorr signing:**
```rust
let sig = secp.sign_schnorr(
            &Message::from_slice(&hash).unwrap_throw(),
            &keypair,
);
```
**FROST Signing (Fails):** 
When using FROST to sign the transaction, we provide the same hash as the message parameter for the SigningPackage. 

```rust
let signing_package =
                frost::SigningPackage::new(signing_state.signing_commitments.clone(), hash);
```

The process creates an aggregated signature, but broadcasting the transaction results in the above error.

Question:

Given the transition from successful individual signing to the failure of the FROST aggregated signature, is there a specific aspect of message handling or signature aggregation with FROST that needs to be adjusted for these transactions? Any insights or suggestions on resolving the "Invalid Schnorr signature" error would be greatly appreciated.

Thank you!

zebra-lucky:
> I'm encountering an issue when broadcasting a PSBT transaction signed using FROST for aggregated Schnorr signatures. While individual Schnorr signatures work correctly, the aggregated signature fails with the following error:

Hello!

Need to clarify details. On the first commits to the this PR there was
a problems with combinations of pubkey/tweaked pubkey,
which lead to invalid sig. Currently it's seems to fixed with @mimoo help.

Do you use the lates commit or frost-secp256k1-tr b3604963317a93ab8398e674d02c9489a7697a6a?

There is some testing code on the testnet to check how it's works:
https://github.com/zebra-lucky/sign-tx-frost

zebra-lucky:
> Also I'm wondering if I need the _tr library for the signing process I'm doing, or if it would be enough to just use the frost_secp256k1 library?
frost-secp256k1 is a general on a curve, frost-secp256k1-tr add some specifics from BIP340/341

frost-secp256k1-tr/src/lib.rs
zebra-lucky: > Thank you. I'll look to code in zkbitcoin in a few days.

Only looking in the current day...

conduition:
> I'm encountering an issue when broadcasting a PSBT transaction signed using FROST for aggregated Schnorr signatures. While individual Schnorr signatures work correctly, the aggregated signature fails with the following error:

@Polybius93 The standard `frost-secp256k1` crate isn't compatible with BIP340 taproot signatures by default. The reason is twofold:

- FROST uses differently namespaced hash functions to determine whether signatures are valid. Also FROST encodes the nonce as a compressed point instead of a BIP340 x-only point when hashing it for the challenge. This PR [addresses that here](https://github.com/ZcashFoundation/frost/pull/584/files#diff-8e709a4a3d634c18a9a3fe98db035be73eb028c914e228d5f41349d0d22ab5b1R275) and [here](https://github.com/ZcashFoundation/frost/pull/584/files#diff-8e709a4a3d634c18a9a3fe98db035be73eb028c914e228d5f41349d0d22ab5b1R321).
- BIP340 has [a lot of baked-in logic](https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki#default-signing) which was added to allow 32-byte compact x-only pubkeys, and 64-byte compact signatures without malleability. It requires signing code to constantly check _'does the point generated by this scalar have an even Y coordinate? if not, negate the scalar.'_ FROST doesn't do this by default, hence why this PR has lots of `if taproot { /* do negation stuff */ } else { ... }` statements. 

To be able to sign taproot transactions, this PR will be needed

Polybius93:
Thank you for the answer @conduition , I appreciate it!

Polybius93:
Hey! Thank you for your answer @zebra-lucky ! Yes, I am using the latest commit. I was looking into your testing code, thank you for sharing it! Currently we are still trying to figure out how to make a valid signature for our use case. I'll get back to you if we have some answers! Thank you!

0xBEEFCAF3:
@Polybius93 
Something else to keep in mind. Looks like you are doing a taproot script spend.
```rust
let hash = SighashCache::new(&psbt.unsigned_tx.clone())
            .taproot_script_spend_signature_hash(
                0,
                &Prevouts::All(&[tx_out.clone()]),
                ScriptPath::with_defaults(&multisig_script),
                SchnorrSighashType::Default,
            )
            .unwrap_throw();
```
 Anyone else feel free to correct me if I am wrong, but this crate hardcodes an empty taptweak. Meaning if the FROST agg key is your internal taproot key then you cannot spend from an alternative tapscript. It is however possible to use the FROST agg key in a tapscript with `OP_CHECKSIG`.

sosaucily:
Thanks @0xBEEFCAF3 , Yep I think that's the problem, @Polybius93 FYI. We're going to look into either adding support for script spend into this branch, or perhaps forking off this fork and building it in there!

conduition:
PR opened here: https://github.com/zebra-lucky/frost/pull/2

Its purpose is to improve the overall readability and 'correctness' of the naming conventions in this PR. It renames all of the methods to be more generally applicable to any ciphersuite, and removes the `is_taproot_compat` method in favor of stubbing the methods to return sensible defaults under any ciphersuite. 

I also added more rustdoc comments to make the public changes on this PR more developer-friendly.

zebra-lucky:
> PR opened here: [zebra-lucky#2](https://github.com/zebra-lucky/frost/pull/2)

Thank you! I'll review it now.

zebra-lucky:
Can not test today with `sign-tx-frost`, created issue https://github.com/zebra-lucky/sign-tx-frost/issues/1

Try to fix it tomorrow.

zebra-lucky:
- Fixed docstrings as mentioned in: https://github.com/zebra-lucky/frost/pull/2#discussion_r1498551101
- Fixed `sign-tx-frost` with use of `rust-bitcoin`, sign/verify is ok

Possibly need to rebase on current branch, will do it in a day.

conduition:
@zebra-lucky I'm noticing a footgun here. The `VerifyingKey` by default is un-tweaked, so a naive caller won't be able to verify signatures unless they manually tweak the verifying key themselves with an empty tweak. [Here's how your code is doing it](https://github.com/zebra-lucky/sign-tx-frost/blob/21092dd516cf1b020c24ea2c287413fe2255daf1/src/main.rs#L382-L386). For those unfamiliar with taproot, this is pretty arcane. 

We should provide some way to convert the untweaked verifying key into a tweaked one, or and hopefully include some way to configure how and whether to tweak signatures. That way this PR can be used not just for key-spending, but also for committing to arbitrary taproot tweaks. 

I'm gonna have a think about this and get back to you next week with some suggestions

zebra-lucky:
> I'm gonna have a think about this and get back to you next week with some suggestions

Thank you! I'll reread BIP341 on script spend variants, and look to early mentioned https://github.com/sigma0-xyz/zkbitcoin for examples.

zebra-lucky:
Rebased `add-secp256k1-tr` branch on current `main` branch

conduition:
@zebra-lucky Unfortunately I don't think I'll have time to contribute code to this in the next few weeks. I hope you'll hear out my suggestions below. i'm not a maintainer of this crate or anyone special, just trying to help out and hopefully move this important PR forward :slightly_smiling_face: 

I think for taproot support to be really useful to people, we should add the ability to tweak each signing session with a custom merkle root (or to omit the tweak entirely, which might be desirable in some scenarios)

I would suggest adding a new field such as `tweak` or `merkle_root` to the `SigningPackage` struct type, or maybe even to the `KeyPackage` struct type. For non-taproot ciphersuites this could default to being empty/unused.

On the taproot crate, this field could be manipulated by taproot signers to configure their signing group to use a specific merkle root or a specific tweak value. If everyone agrees on the same tweak, then signatures would validate under a tweaked verifying key, which could be returned using a new method defined in the `frost-secp256k1-tr` crate.

Let me know what you think of this, open to alternative suggestions for a cleaner API

conduition:
I couldn't help myself. 

@zebra-lucky [I made a PR into your branch](https://github.com/zebra-lucky/frost/pull/3) which adds 
- the ability for signers to commit to a specific tapscript merkle root at signing time
- the ability to verify a signature commits to a specific tapscript merkle root
- the ability to opt out of tweaking
- the ability to export the tweaked `VerifyingKey`

## Custom Tweaking
Instead of a message byte slice, the caller passes a `SigningTarget` when constructing the `SigningPackage`. This encapsulates both the message to sign, and the tapscript merkle root:

```rust
use frost_secp256k1_tr::{SigningPackage, SigningTarget, SigningParameters};

let merkle_root = vec![0u8; 32];

let sig_params = SigningParameters {
    tapscript_merkle_root: Some(merkle_root),
};
let pkg = SigningPackage::new(
    commitments,
    SigningTarget::new(b"a message to sign", sig_params)
);
```

To commit to an unspendable taproot tweak, set `merkle_root = vec![]` explicitly.

## No Tweak

To use FROST without taproot tweaking, simply sign messages the way you would with any other ciphersuite.

```rust
let pkg = SigningPackage::new(commitments, b"a message to sign");
```

This uses the FROST group key (taproot internal key) directly as the taproot output key during signing with no tweak.

The above is how most downstream callers will probably want to use BIP340, and this keeps the API identical to `master` while also allowing lots of flexibility for more complex taproot use-cases. Under the hood, it is equivalent to explicitly opting-out of taproot tweaking by setting `tapscript_merkle_root` to `None`, like so: 

```rust
let sig_params = SigningParameters {
    tapscript_merkle_root: None,
};
let pkg = SigningPackage::new(
    commitments,
    SigningTarget::new(b"a message to sign", sig_params)
);
```

This is still safe, because the DKG protocol protects the FROST group key from being influenced by rogue keys. However, if 3rd party observers need to verify the script path is unspendable, then an explicit empty merkle root would be needed.


## Verifying
When verifying signatures, you can use the untweaked `VerifyingKey` and provide the same `SigningTarget` used to create the signature (including the same merkle root). 

```rust
let sig_params = SigningParameters {
    tapscript_merkle_root: Some(merkle_root),
};
verifying_key.verify(
    SigningTarget::new(b"a message to sign", sig_params),
    signature,
)?;
```

Alternatively, any FROST group member can provide the verifier with a tweaked `VerifyingKey`, and the verifier should then be able to check the message signature without learning the taproot tweak used to produce the signature.

```rust
verifying_key.effective_key(&sig_params).verify(
    b"a message to sign",
    signature,
)?;
```

zebra-lucky:
> @zebra-lucky Unfortunately I don't think I'll have time to contribute code to this in the next few weeks. I hope you'll hear out my suggestions below. i'm not a maintainer of this crate or anyone special, just trying to help out and hopefully move this important PR forward 🙂

Thank you, @conduition, your PR with code refactoring was very helpful!
I'll do suggested changes in a day or two.

zebra-lucky:
Sorry @conduition, I missed the mention of the new PR #3 yesterday, thank you very much. Today I will explore the code and docs.

zebra-lucky:
Merged https://github.com/zebra-lucky/frost/pull/3/ from @conduition.
Looks really good, and I very thankful for help.
Currently I has very low productivity.

zebra-lucky:
Tomorrow I'll review https://github.com/zebra-lucky/sign-tx-frost code and test usage changes from https://github.com/zebra-lucky/frost/pull/3.

mimoo:
sry to spam here but do you have a telegram @zebra-lucky ? I'm @cryptodavidw

frost-core/src/tests/vectors.rs
```diff
@@ -311,6 +311,10 @@ pub fn check_sign_with_test_vectors<C: Ciphersuite>(json_vectors: &Value) {
 
     // Check that the generated signature matches the test vector signature
     let group_signature = group_signature_result.unwrap();
+    assert_eq!(
+        hex::encode(group_signature.serialize()),
+        hex::encode(signature_bytes.clone()),
+    );
```
conduition, line 317: @zebra-lucky My bad, i think this was a left-over from my debugging phase. You can remove this.

conduition:
My pleasure @zebra-lucky, I'm happy to help. This PR would be very useful for a lot of people. I think it's close to being ready for review soon, is there anything you'd like to tidy up?

zebra-lucky:
> sry to spam here but do you have a telegram @zebra-lucky ? ...

I'm sorry, you can use zebra.lucky at gmail to communicate.

frost-core/src/tests/vectors.rs
zebra-lucky: Fixed with 1c085ba

zebra-lucky:
> My pleasure @zebra-lucky, I'm happy to help. This PR would be very useful for a lot of people. I think it's close to being ready for review soon, is there anything you'd like to tidy up?

I think only add changes to `sign-tx-frost` to use some scripts and check on real transactions on the testnet.
Possibly in a few days. Low productivity currently.

MatthewLM:
@zebra-lucky Is the `tapscript_merkle_root` not supposed to work yet? I get an invalid signature error when I try to use it. It appears that the verification in the aggregation function doesn't apply the tweak to the public key:

```
    // Verify the aggregate signature
    let verification_result = pubkeys
        .verifying_key
        .verify(signing_package.message(), &signature);
```

MatthewLM:
@zebra-lucky I changed the verify function to accept the `SigningTarget` directly:

```
    pub fn verify(
        &self,
        sig_target: &SigningTarget<C>,
        signature: &Signature<C>,
    ) -> Result<(), Error<C>> {
        C::verify_signature(&sig_target, signature, self)
    }
```

```
    let verification_result = pubkeys
        .verifying_key
        .verify(signing_package.sig_target(), &signature);
```

And this works, though it breaks passing a message directly into the verify function, so I'm not sure how you'd want to handle that.

conduition:
@MatthewLM Good catch. Looks like we don't have any unit test coverage for tweaked signatures yet. I'll fix that alongside the bug you found there.

conduition:
Integration tests and a fix for the bug pointed out by @MatthewLM have been submitted to this PR in https://github.com/zebra-lucky/frost/pull/4

zebra-lucky:
Thank you, @conduition for your help.
Your help in this code was invaluable.

I'm sorry for a pause, but i'm ill currently.
But i'll rewiew PR for a hours....

MatthewLM:
@conduition @zebra-lucky I've not experienced any further issues using 20c2c98, thank you for the fix. I've been able to create successful Taproot key-path and script-path spends. I look forward to seeing this finished and merged in.

Given Taproot support, I'm curious about deriving unhardened BIP32 keys. Clearly hardened derivation is not possible, except perhaps by modifying the scheme somehow. A straight-forward method for unhardened derivation might be to first derive a master chaincode from the participant's public key shares. Then the same tweaks can be applied to the group key, public share and private shares.

I wonder if there are any security implications to deriving BIP32 keys this way though. I don't see how there would be but I know there might be some nuance that I'm unaware of.

conduition:
@MatthewLM Intriguing idea. I hadn't considered BIP32 support. Yes, hardened derivation wouldn't be possible without some special DKG-like ritual, which would probably be more trouble than it's worth (At that point you could just run the DKG again to make a fresh FROST key).

> A straight-forward method for unhardened derivation might be to first derive a master chaincode from the participant's public key shares. Then the same tweaks can be applied to the group key, public share and private shares.

This approach should be safe. The chain code is public data (public in the sense of "it reveals nothing that would help an observer forge signatures"), so deriving it from the FROST group's public verification shares should be acceptable. Group members would not be concerned that the chaincode was chosen unfairly, because each member is presumed to have participated in the DKG process which produced the verification shares (and thus the chain code). 

There are down-sides to that approach though. If the full set of verification shares must be exposed for any reason (e.g to prove each signer's membership in the group), this would also result in exposing the full extended public key for the group. 

Also, this approach would not be practically compatible with [Verifiable Secret Resharing (VSR)](https://github.com/ZcashFoundation/frost/issues/519) or a [Repairable Threshold Scheme](https://docs.rs/frost-core/latest/frost_core/keys/repairable/index.html), because those techniques involve adding or otherwise changing the set of shares. Doing so would also change the chain code derived from those shares, and thus the group extended pubkey would need to change. I imagine many use cases would prefer that not to happen. 

I have a few alternative suggestions for how to combine FROST and BIP32, but i think for the sake of scope-reduction we should keep this PR limited to a single static group key. Additive tweaks on signing shares are easy for downstream callers to add themselves, by using the `internals` flag to access and modify shares directly. This PR is a lower-level prereq for taproot compatibility, so we should avoid biting off more than we (or the repository maintainers) can chew here.

MatthewLM:
@conduition Thanks for your response. I agree with what you say. The risk of leaking all of the public shares could perhaps be mitigated if membership can be proved with zero-knowledge. Deriving the chaincode from the public shares would be trivial but I agree in some (perhaps many) cases it would not be desirable. A chain-code could be separately generated between all participants through an extension of the DKG.

I agree that it complicates this PR, so maybe I could create a discussion on this repo for BIP32 considerations?

conduition:
> I agree that it complicates this PR, so maybe I could create a discussion on this repo for BIP32 considerations?

I think that would be a swell idea :smile:

MatthewLM:
@conduition I started a discussion here: https://github.com/ZcashFoundation/frost/discussions/636

conduition:
@zebra-lucky I realized we were still serializing taproot signatures as 65-byte arrays with compressed R points. To be compatible with BIP340, callers would need to serialize and manually trim that leading byte. For usability, i have submitted a PR which fixes this, ensuring signatures are always computed with even R points, and serialized as BIP340-compatible 64-byte arrays. https://github.com/zebra-lucky/frost/pull/5

zebra-lucky:
I'm sorry for a delayed reaction.
Lookin to code changes. Need to run tests.

conduition:
Not delayed at all, man. Thanks very much for your quick attention :smile:

ArchSpaceman:
Hey all !

Thank you for your contributions to this PR. I've opened a new PR  to re-export frost-core as it is currently abstracted in a way that prevents us from verifying individual signature shares. [https://github.com/zebra-lucky/frost/pull/6](url)
This change is proposed to ensure that the internal functions needed are accessible for individual signature share verification.

We would love if you take a moment to review the PR's changes, and provide any feedback !

sosaucily:
Hi @zebra-lucky and @conduition, I'm just following up on this thread as it's been quiet for a while. I see This library released a 2.0.0-rc recently, and I wonder if this feature might be part of the full 2.0.0 release. We're not really blocked as we're working off the branch, but it would be great to get merged into the upstream code. 

Thanks!

mpguerra:
> I see This library released a 2.0.0-rc recently, and I wonder if this feature might be part of the full 2.0.0 release. 
@conradoplg @natalieesk what do you think?

conduition:
@sosaucily I would love to get this into the 2.0.0 release. Unfortunately i'm not sure when this PR will be merged - I imagine the maintainers of the `frost` repo will probably want to request changes, such as a different API, additional documentation coverage or tests.

i also have an additional feature branch, [`taproot-bip32`](https://github.com/conduition/frost/compare/add-secp256k1-tr..taproot-bip32), based off of `zebra-lucky:add-secp256k1-tr` which adds BIP32 support to the `frost-secp256k1-tr` crate. See [this discussion](https://github.com/ZcashFoundation/frost/discussions/636) for more background. It would be nice to be able to PR that into this repo as well, so that FROST keys can be used in hierarchical and descriptor wallets.

zebra-lucky:
Hi!

At the first I very thankful to the people who gave a ideas on the PR:
- @akhavr on the plan
- @mimoo on the perfect math and graphics
- @conduition on the perfect knowledge of Rust and the bictcoin
- developers of ZcashFoundation on the attention

need to think how to transfer PR code, to not do a new PR

zebra-lucky:
> need to think how to transfer PR code, to not do a new PR
my productivity currently is at very low, I'm sorry

conradoplg:
Thanks everyone, I really appreciate the work that went on this.

I want to get another look into this and I might do some refactorings to avoid changing the frost_core code so much. I think it's unlikely that this will make into the 2.0.0 release, but we'll probably do a 2.1.0 release not much time after that and this might be included then.

Don't worry about the conflicts, I'll solve them.

ChristopherA:
We are hosting our send FROST library implementors workshop on September 18th https://developer.blockchaincommons.com/frost/meeting2/

So far scheduled are some discussion from bitcoin devs on their FROST implementations in C++. One of the critical issues to me are the issues of x-only pubkey compatibility and the tweaked signatures for use on-chain. All of Blockchain Common's own code base has been moving from C++ to Rust and we'd prefer a rust implementation.

Could we get someone from this team to participate, or even present a brief status update / roadmap on this PR? Contact me at ChristopherA@LifeWithAlacrity.com if you are interested.

Thanks!

zebra-lucky:
> @zebra-lucky [It looks like it's not possible to transfer a pull request on github unfortunately](https://stackoverflow.com/questions/66539231/how-to-change-the-owner-of-a-pr-on-github-how-to-commandeer-an-open-github-pr0), we'd have to close this PR and open a new one if we wanted to change the PR owner. Would you prefer I open a new PR from my fork of `frost`? that way you wouldn't need to be involved reviewing/merging stuff anymore

Hi! Make it in your preferred way.

conduition:
I have re-opened this PR here under my ownership: https://github.com/ZcashFoundation/frost/pull/730 - All further changes should be made to that PR's branch. CC @sosaucily

PR #597, opened 2024-01-16
Check num of commitments against min signers

natalieesk:
Throws an error if the number of commitments is incorrect

Updates `.get(0)` to `.first()`


mpguerra:
@natalieesk can you please add an estimate to this one?

PR #601, opened 2024-01-25
frost-core: expose SigningKey::to/from_scalar()

### LINKED ISSUE #497: `SigningKey::{from_scalar, to_scalar}` methods are inconveniently hidden
These methods are gated by an `internals` feature:
```rust
    /// Creates a SigningKey from a scalar.
    #[cfg(feature = "internals")]
    pub fn from_scalar(
        scalar: <<<C as Ciphersuite>::Group as Group>::Field as Field>::Scalar,
    ) -> Self {
        Self { scalar }
    }

    /// Return the underlying scalar.
    #[cfg(feature = "internals")]
    pub fn to_scalar(self) -> <<<C as Ciphersuite>::Group as Group>::Field as Field>::Scalar {
        self.scalar
    }
```
We want to use these methods to bridge between our existing key types and the FROST types; it'd be convenient if the methods were exposed by default.

I don't think there's any meaningful increase in encapsulation, since the secret is already exposed via the serialization methods.

conradoplg:
Closes #497

### DIFF
--- a/frost-core/CHANGELOG.md
+++ b/frost-core/CHANGELOG.md
@@ -4,7 +4,7 @@ Entries are listed in reverse chronological order.
 
 ## Unreleased
 
-* Exposed the `SignignKey::from_scalar()` and `to_scalar()` methods. This
+* Exposed the `SigningKey::from_scalar()` and `to_scalar()` methods. This
   helps interoperability with other implementations.
 
 ## Released
--- a/frost-core/CHANGELOG.md
+++ b/frost-core/CHANGELOG.md
@@ -4,6 +4,8 @@ Entries are listed in reverse chronological order.
 
 ## Unreleased
 
+* Exposed the `SignignKey::from_scalar()` and `to_scalar()` methods. This
+  helps interoperability with other implementations.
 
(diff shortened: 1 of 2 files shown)


frost-core/CHANGELOG.md
```diff
@@ -4,6 +4,8 @@ Entries are listed in reverse chronological order.
 
 ## Unreleased
 
+* Exposed the `SignignKey::from_scalar()` and `to_scalar()` methods. This
```
natalieesk, line 7: typo `SigningKey::from_scalar()`

natalieesk (changes requested):
Just a typo

frost-core/CHANGELOG.md
conradoplg, line 7: Oops, thanks!

```suggestion
* Exposed the `SigningKey::from_scalar()` and `to_scalar()` methods. This
```

conradoplg: Done in 6ba38c3

natalieesk (approved):
Great

PR #621, opened 2024-02-27
add no-std support

### LINKED ISSUE #535: Support no_std
It's likely that somebody will need no_std support at some point so we should support it.

I haven't found a good reference on how to do that (some [basic info](https://docs.rust-embedded.org/book/intro/no-std.html)) but this basically entails having a std feature that is enabled by default, and gating std stuff behind it. See how it's done in reddsa repo.

conradoplg:
I didn't add it to frost-ed448 because Ed448-Goldilocks doesn't support it. I'll eventually create a PR there and we can enable it in frost-ed448 later.

Closes #535 

This is a draft just because it's a breaking change and we need to decide when to merge it. Otherwise it's ready.


drewstone:
In case it's useful I have a `no_std` lib of Ed448: https://github.com/webb-tools/Ed448-Goldilocks

conradoplg:
> In case it's useful I have a `no_std` lib of Ed448: https://github.com/webb-tools/Ed448-Goldilocks

This is great, but we'd prefer if that was incorporated upstream. Did you consider opening a PR there?

natalieesk (approved):
Looks good

natalieesk (approved):

PR #624, opened 2024-02-28
ci: test MSRV and minimal versions; also fix thiserror min version

conradoplg:


.github/workflows/main.yml
```diff
@@ -24,6 +24,20 @@ jobs:
...
           command: build
 
+  build_msrv:
+    name: build with MSRV (1.66.1)
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4.1.1
+      # Re-resolve Cargo.lock with minimal versions
+      - uses: dtolnay/rust-toolchain@nightly
```
gustavovalverde, line 34:
```suggestion
      - uses: dtolnay/rust-toolchain@stable
```

Using a nightly version might break CI from time to time.

gustavovalverde (approved):
I'm approving, but taking the review into consideration would be preferable.

.github/workflows/main.yml
conradoplg: Unfortunately we can't since `cargo update -Z minimal-versions` is [still unstable](https://doc.rust-lang.org/cargo/reference/unstable.html#minimal-versions), I think it's an acceptable risk

gustavovalverde (approved):
LGTM

PR #626, opened 2024-03-01
core: remove MulAssign impl for Scalar

### LINKED ISSUE #625: Breakage in future Rust version
Hello there, unfortunately `frost-core` uses an impl which should result in a coherence error, but due to an oversight in the existing compiler implementation, it has not been detected up until now. This impl must be allowed as it may overlap.

https://github.com/ZcashFoundation/frost/blob/6e5aff8342fa4f0fe99720931921f94255e64900/frost-core/src/identifier.rs#L152-L159

Scalar is defined as the associated type `<<<C as Ciphersuite>::Group as Group>::Field as Field>::Scalar`. Were a user to implement `Field` for their own type they could choose an associated type which already implements `MulAssign<Identifier<Itself>>`, causing an overlap. This is https://github.com/rust-lang/rust/issues/99554.

We're currently working on a fix in https://github.com/rust-lang/rust/pull/117164 and will likely land it as a future compat warning for now. This will then change to a hard error in the medium term. I can't tell how to best avoid this error? It may potentially be easiest to remove that impl entirely?

I apologize for the inconvenience and thank you for your understanding

conradoplg:
This is a breaking change, so we might want to wait a bit before merging it. It's unlikely it will cause problems in the near future since it's unlikely someone is using MulAssign.

I created as a draft until we decide when we merge it, but it's ready for review.

Closes #625


natalieesk (approved):
Looks good

PR #665, opened 2024-05-29
Add refresh shares with dealer functionality

### LINKED ISSUE #245: Feature: remove signers or refresh shares with trusted dealer

natalieesk:
This closes #245


frost-core/src/tests/refresh.rs
```diff
@@ -0,0 +1,128 @@
...
+        frost::keys::IdentifierList::Default,
+        &mut rng,
+    )
+    .unwrap();
+
+    // Try to refresh shares
+    // Signer 2 will be removed and Signers 1, 3, 4 & 5 will remain
+
+    // Rerun key generation
+
```
conradoplg, line 126: Stale comments that can be removed?

```suggestion
```

frost-core/src/tests/refresh.rs
```diff
@@ -0,0 +1,128 @@
...
+        BTreeMap<Identifier<C>, SecretShare<C>>,
+        PublicKeyPackage<C>,
+    ) = frost::keys::split(
+        &key,
+        max_signers,
+        min_signers,
+        frost::keys::IdentifierList::Default,
+        &mut rng,
+    )
+    .unwrap();
```
conradoplg, line 120: Any particular reason for not just calling `generate_with_dealer()` like in the previous test?

frost-core/src/keys/refresh.rs
```diff
@@ -0,0 +1,119 @@
...
+
+/// Refreshes shares using a trusted dealer
+pub fn refresh_shares_with_dealer<C: Ciphersuite, R: RngCore + CryptoRng>(
+    old_shares: BTreeMap<Identifier<C>, SecretShare<C>>,
+    old_pub_key_package: PublicKeyPackage<C>,
+    max_signers: u16,
+    min_signers: u16,
+    identifiers: &[Identifier<C>],
+    rng: &mut R,
+) -> Result<(BTreeMap<Identifier<C>, SecretShare<C>>, PublicKeyPackage<C>), Error<C>> {
```
conradoplg, line 27: I think we'll need to refactor the API. Consider the regular trusted dealer: the dealer generates a bunch of `SecretShares` that are sent to the participants which convert them to `KeyPackage`s.

If they want to refresh their shares, with this API, they would need to:

- Sent their KeyPackages back to the Coordinator, which would then need to convert them back to SecretShares
- Call `refresh_shares_with_dealer()`
- Send the refreshed `SecretShares` to the participants

An easier flow would be:

- The Coordinator calls something like `generate_refresh_shares_with_dealer()` (name ideas welcome) which is basically just `split()` with a zero key
- The Coordinator sends the zero-key `SecretShares` to participants
- Each participant would them call a new e.g. `refresh_key_package()` with their current `KeyPackage` and the zero-key `SecretPackage`, add them with the logic you already wrote and return the refreshed `KeyPackage`

Another upside is that `refresh_key_package()` would also be used by the DKG refresh share.

conradoplg (changes requested):
I added some comments, the main one is about possibly refactoring the API.

frost-core/src/keys/refresh.rs
```diff
@@ -0,0 +1,106 @@
...
+        verifying_shares.insert(share.identifier, signer_public);
+    }
+
+    let pub_key_package = PublicKeyPackage::<C> {
+        header: old_pub_key_package.header,
+        verifying_shares,
+        verifying_key: old_pub_key_package.verifying_key,
+    };
+
+    Ok((zero_shares, pub_key_package))
```
conradoplg, line 64: As I mentioned in Slack, there will be an issue here: the first commitment will be the identity, and will fail to be serialized. We could remove it from the commitment vector and re-add it in `refresh_share`.

Also, at first, I thought that the right `verifying_key` to return it would be matching verifying key for the zero key. But that would cause the same issue. So we can keep using the old verifying key.

Additionally, most of this function could be replaced by calling `split()` since the logic is the same.

frost-core/src/tests/refresh.rs
```diff
@@ -0,0 +1,142 @@
...
+
+    let (zero_shares, new_pub_key_package) = calculate_zero_key(
+        pub_key_package,
+        NEW_MAX_SIGNERS,
+        min_signers,
+        &remaining_ids,
+        &mut rng,
+    )
+    .unwrap();
+
```
conradoplg, line 60: At this point we can do a serialization round trip to make sure we are not serializing the identity.

conradoplg (changes requested):
We'll need to handle the serialization issue, avoiding returning the identity element

frost-core/src/keys/refresh.rs
```diff
@@ -0,0 +1,106 @@
...
+}
+
+/// Each participant refreshed their shares
+pub fn refresh_share<C: Ciphersuite>(
+    zero_share: SecretShare<C>,
+    current_share: &SecretShare<C>,
```
conradoplg, line 70: Participants hold onto KeyPackages, so here `curretn_share` should be a `KeyPackage`

conradoplg (changes requested):
Looks good, thanks! It just needs fixing CI which requires just changing the import of `Vec`.

I'm just not sure about the `calculate_zero_key` name. I don't think the user should care about the fact that involves the zero key. Maybe something like `compute_refreshing_shares`?

There's some places where I think we can make the code a bit clearer but I think I'll create a PR after this one merges and we can then discuss it.

book/src/tutorial/refreshing-shares.md
```diff
@@ -0,0 +1,28 @@
...
+The diagram below shows the refresh share process. Dashed lines
+represent data being sent through an [authenticated and confidential communication
+channel](https://frost.zfnd.org/terminology.html#peer-to-peer-channel).
+
+<!-- ![Diagram of Refreshing shares, illustrating what is explained in the text](refreshing.png) -->
+
+The Trusted Dealer needs to first run `compute_refreshing_shares` where the new SecretShares are generated and then verified.
+This is done with
+[`KeyPackage::try_from()`](https://docs.rs/frost-core/latest/frost_core/frost/keys/struct.KeyPackage.html#method.try_from):
+`caluclate_zero_key` returns a new SecretShare and PublicKeyPackage
```
conradoplg, line 12:
```suggestion
`compute_refreshing_shares` returns a new SecretShare and PublicKeyPackage
```

conradoplg (approved):
Looks good, thanks!

PR #671, opened 2024-06-08
return error when deserializing the identity

### LINKED ISSUE #667: Reject identity in Group::serialize
In current RFC document 3.1, page 10: 
> SerializeElement(A): Maps an Element A to a canonical byte array
buf of fixed length Ne. This function raises an error if A is the
identity element of the group.

Serialize a Group identity element should raise error.

Related PR:
https://github.com/cfrg/draft-irtf-cfrg-frost/pull/306
https://github.com/ZcashFoundation/frost/pull/145

conradoplg:
Closes #667 

To keep serde working, I switched to another approach which is actually much simpler. Instead of using `serde(try_from/into)` (which I no longer couldn't since there is no `try_into`) I simply created a `SerializedElement` type which does define serde encoding and used it in place of a raw `Element`. This allows the struct that contains it to support serde by simply deriving it. It does introduce some friction by requiring using e.g. `val.0 .0` but that's internal (and can be avoided with getters). We should probably use the same approach for Scalars, but we can do that in a separate PR.

### DIFF
--- a/frost-core/CHANGELOG.md
+++ b/frost-core/CHANGELOG.md
@@ -4,6 +4,23 @@ Entries are listed in reverse chronological order.
 
 ## Unreleased
 
+* Changed the `deserialize()` function of Elements and structs containing
+  Elements to return an error if the element is the identity. This is a
+  requirement in the FROST specification that wasn't being followed. We are not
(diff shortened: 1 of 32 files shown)


natalieesk (approved):
This is great

PR #674, opened 2024-06-12
Make serialization methods consistent

### LINKED ISSUE #661: Make serialization methods consistent
There are currently 3 types of (de)serialization functions:

1. `fn deserialize(bytes: <Something>::Serialization)` which is used for Scalar, Field, and Signature types and wrappers
1. `fn deserialize(bytes: &<Something>::Serialization)` same as the previous, but with a reference
1. `fn deserialize(bytes: &[u8])` which is used for all other structs

The obvious inconsistency is between 1 and 2. I was changing all of them to take a reference (option 2) but thinking about it, I wonder if we should make them all follow option 3.

The `<Something>::Serialization` seems like an implementation detail which users shouldn't need to handle. Currently, to deserialize e.g. an identifier you need something like

```rust
        frost_core::Identifier::<C>::deserialize(
            &bytes
                .try_into()
                .map_err(|_| frost_core::Error::<C>::DeserializationError)?,
        )
```

which is super annoying. The user needs to lookup what the `<<C::Group as Group>::Field as Field>::Serialization` type is, and figure out that they have to use `TryFrom<Vec<u8>>`.

If we changed to option 3 in the public API (while still using `<<C::Group as Group>::Field as Field>::Serialization`  internally) then they just have to pass a `&[u8]` which is more straightforward.

conradoplg:
Depends on #671 

Closes #661

### DIFF
--- a/frost-core/CHANGELOG.md
+++ b/frost-core/CHANGELOG.md
@@ -17,6 +17,18 @@ Entries are listed in reverse chronological order.
     `VerifiableSecretSharingCommitment::serialize()`,
     `NonceCommitment::serialize()`, `Signature::serialize()`,
     `VerifyingKey::serialize()` can now all return an error.
+* Changed the `serialize()` and `deserialize()` methods of all Scalar- and
+  Element-wrapping structs; instead of taking or returning a
+  `Field::Serialization` or `Element::Serialization` traits (which are usually
+  defined by ciphersuites as arrays of specific sizes), they simply respectively
+  take `&[u8]` and return `Vec<u8>`, exactly as the other structs, which should
+  greatly simplify non-serde serialization code. You can port existing code with
+  e.g. `x.serialize().as_ref()` -> `x.serialize()` and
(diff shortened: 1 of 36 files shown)


frost-core/CHANGELOG.md
```diff
@@ -17,6 +17,18 @@ Entries are listed in reverse chronological order.
     `VerifiableSecretSharingCommitment::serialize()`,
     `NonceCommitment::serialize()`, `Signature::serialize()`,
     `VerifyingKey::serialize()` can now all return an error.
+* Changed the `serialize()` and `deserialize()` methods of all Scalar- and
+  Element-wrapping structs; instead of taking or returning a
+  `Field::Serialization` or `Element::Serialization` traits (which are usually
```
natalieesk, line 22: Typo, should be "trait" not "traits"

conradoplg:
```suggestion
  `Field::Serialization` or `Element::Serialization` trait (which are usually
```

conradoplg: Fixed

natalieesk (approved):
Looks good

PR #689, opened 2024-06-19
build and test each crate separately

conradoplg:
`cargo test -p frost-core` was failing because the benchmark was failing to build, which in turn was being caused by the `criterion` dependency not being included when compiling without features.

This was not being caught because on CI we just use `cargo test`, which merges all features and would end up including `criterion`.

This was also not being caught by the all-features test because we were only doing `build` (which does not build benchmarks) and not `test` (which does build benchmarks).

This fixes the issue, and adds test-all-features to CI, which already works on each crate separately.


natalieesk (approved):
Looks good

PR #699, opened 2024-07-04
docs: update demo section per new ciphersuite selection

conradoplg:


mpguerra (approved):
Checked that old commands no longer work and new commands work

PR #712, opened 2024-08-04
chore(docs): fix clippy warnings in examples

StackOverflowExcept1on:
- simplify range expression
- use `key_package` variable


conradoplg (approved):
Thanks!

PR #714, opened 2024-08-07
Add wget of zcash-params 

dismad:
Add wget for zcash-params to make it easier to download needed files in proper location.


conradoplg:
This will be not longer needed after https://github.com/ZcashFoundation/frost-zcash-demo/issues/217 merges because params will be embedded, but let's leave this open until me ensure it is truly not needed.

PR #721, opened 2024-08-13
docs: update demo docs for user registration

conradoplg:
This should merge only after https://github.com/ZcashFoundation/frost-zcash-demo/pull/252 gets merged

### DIFF
--- a/book/src/zcash/ywallet-demo.md
+++ b/book/src/zcash/ywallet-demo.md
@@ -141,13 +141,14 @@ guide we will need two. In a new terminal, run the following command for user
 curl --data-binary '{"username": "alice", "password": "foobar10", "pubkey": ""}' -H 'Content-Type: application/json' http://127.0.0.1:2744/register
 ```
 
-It will output "null". Also register user "bob":
+It will output "null". (The "pubkey" parameter is not used currently and should
+be empty.) Also register user "bob":
 
 ```
 curl --data-binary '{"username": "bob", "password": "foobar10", "pubkey": ""}' -H 'Content-Type: application/json' http://127.0.0.1:2744/register
 ```
 
-You only need to do this once, in case you want to sign more than one
+You only need to do this once, even if you want to sign more than one
 transaction. If for some reason you want to start over, close the server and
 delete the `db.sqlite` file.
 
@@ -204,6 +205,7 @@ Partcipant too!)
 In the fourth terminal, for Participant 2, run the following:
 
 ```
+export PW=foobar10
 cargo run --bin participant -- -C redpallas --http --key-package key-package-2.json -u bob -w PW
 ```
 
--- a/book/src/zcash/ywallet-demo.md
+++ b/book/src/zcash/ywallet-demo.md
@@ -117,6 +117,9 @@ cargo run --release -- sign --tx-plan <tx_plan_path> --ufvk <ufvk> -o <tx_signed
 
 The program will print a SIGHASH and a Randomizer.
 
+
+### Running the server
+
 Now you will need to simulate two participants and a Coordinator to sign the
 transaction, and the FROST server that handles communications between them.
 It's probably easier to open four terminals.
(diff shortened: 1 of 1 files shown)


book/src/zcash/ywallet-demo.md
```diff
@@ -128,36 +131,51 @@ run):
...
-was run):
+### Registering users
+
+In order to interact with the server, you will need to register users. For this
+guide we will need two. In a new terminal, run the following command for user
+"alice" (replace the password if you want):
 
 ```
-cargo run --bin coordinator -- -C redpallas --http -r -
+curl --data-binary '{"username": "alice", "password": "foobar10", "pubkey": ""}' -H 'Content-Type: application/json' http://127.0.0.1:2744/register
```
mpguerra, line 141: MIght be good to explain what the pubkey parameter is about and why it's an empty string here

book/src/zcash/ywallet-demo.md
```diff
@@ -128,36 +131,51 @@ run):
...
+It will output "null". Also register user "bob":
 
-- It should read the public key package from `public-key-package.json`.
-- Type `2` for the number of participants.
-- Copy the session ID that is printed.
+```
+curl --data-binary '{"username": "bob", "password": "foobar10", "pubkey": ""}' -H 'Content-Type: application/json' http://127.0.0.1:2744/register
+```
 
+You only need to do this once, in case you want to sign more than one
```
mpguerra, line 150: This sentence is a bit confusing

mpguerra, line 150:
```suggestion
You only need to do this once, even if you want to sign more than one
```

book/src/zcash/ywallet-demo.md
```diff
@@ -128,36 +131,51 @@ run):
...
 ```
 
-Go back to the Coordinator CLI:
+We will use "alice" as the Coordinator, so change the value next to `export PW=`
+if you used another password when registering "alice".
 
+And then:
+
+- It should read the public key package from `public-key-package.json`.
 - When prompted for the message to be signed, paste the SIGHASH printed by the
```
mpguerra, line 179: Here it's waiting for participants to send their commitments first, maybe we should add this before mentioning the prompt for the message to be signed

book/src/zcash/ywallet-demo.md
```diff
@@ -169,9 +187,31 @@ If you prefer to pass the randomizer as a file by using the `--randomizer`
...
+
+(We are using "alice" again. There's nothing stopping a Coordinator from being a
+Partcipant too!)
+
+### Participant 2 (bob)
+
+In the fourth terminal, for Participant 2, run the following:
+
+```
+cargo run --bin participant -- -C redpallas --http --key-package key-package-2.json -u bob -w PW
```
mpguerra, line 207: Here it prompted me to set the password. Should we add `export PW=foobar10`  to this command also or will it depend on how people have configured things in their terminals?

mpguerra (commented):
Looks pretty good, I think the order of actions to be taken by users could be improved a bit, e.g you need to run the participant commands to generate and send the commitments before  you can paste the SIGHASH for the message to be signed. 
Also, I got an `InvalidExternalSignature` error when pasting the group signature into the sign command but I think this was probably due to user error.

book/src/zcash/ywallet-demo.md
conradoplg: Done in 412dd015

book/src/zcash/ywallet-demo.md
conradoplg: Ah I see, I ended up using https://github.com/ZcashFoundation/frost-zcash-demo/pull/265 to write these docs, and the flow changed there. Since that's already approved I think it's OK, we can merge the docs after both that and the user registration PR merge.

book/src/zcash/ywallet-demo.md
conradoplg: Good catch, fixed in 412dd015

conradoplg (commented):
Thanks! I re-tested the flow and it seems the work, the invalid signature error was probably caused by not using the same keys that were used to generate the address (this could happen if you generated them again after you generated the address)

book/src/zcash/ywallet-demo.md
conradoplg: Fixed in 412dd015

conradoplg: Fixed in 412dd015

mpguerra (approved):
Thank you!

PR #728, opened 2024-09-12
allow identifying culprit in error returned by dkg::part3()

### LINKED ISSUE #725: Returns culprit for `Error::InvalidSecretShare`
The `culprit()` is None for `InvalidSecretShare`:
https://github.com/ZcashFoundation/frost/blob/dcf17732f791cd5c69aeed5bb4ff60d019a57ee8/frost-core/src/error.rs#L128-L136

This is not expected behaviour because this error can be returned in DKG round 3
https://github.com/ZcashFoundation/frost/blob/dcf17732f791cd5c69aeed5bb4ff60d019a57ee8/frost-core/src/keys/dkg.rs#L525-L528

So, because this error can be caused by other participant, I think culprit must be added to `InvalidSecretShare`.

conradoplg:
Closes #725

### DIFF
--- a/frost-core/CHANGELOG.md
+++ b/frost-core/CHANGELOG.md
@@ -10,6 +10,9 @@ Entries are listed in reverse chronological order.
   but it's likely to not require any code changes since most ciphersuite
   implementations are probably just empty structs. The bound makes it possible
   to use `frost_core::Error<C>` in `Box<dyn std::error::Error>`.
+* It is now possible to identify the culprit in `frost_core::keys::dkg::part3()`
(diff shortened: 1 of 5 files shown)


StackOverflowExcept1on:
I think it would also be useful to split `part3` function into several parts. For example, one part takes `max_signers`, `identifier`, `round1_packages`, `round2_packages`. This way, some independent party (like a smart contract) can find culprit, and other participants can locally calculate their secret shares.

natalieesk (approved):
Looks good

Issue #688, opened 2024-06-19
Add checks for serialization round trips during tests


conradoplg:
During the refresh shares work, there was a bug that wasn't showing up on tests because it only happened when serializing the structs generated by the refresh shares functionality.

It would be good to add serialization roundtrips (serialize then deserialize) in the generic FROST tests, to make sure things that need to be serialized can actually be serialized.

PR #730, opened 2024-09-18
Add frost-secp256k1-tr crate (BIP340/BIP341) [moved]

conduition:
This PR is a re-opening of https://github.com/ZcashFoundation/frost/pull/584 to resolve PR ownership problems. I'm opening this so [zebra-lucky (who is short on time) doesn't need to be the middle-man merging my commits into their PR anymore](https://github.com/ZcashFoundation/frost/pull/584#issuecomment-2215405681). 

-------------

This PR adds a new ciphersuite crate `frost-secp256k1-tr`, which complies with [BIP340 Schnorr Signatures on Bitcoin](https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki) and is compatible with [BIP341](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki) tweaks. This crate can be used to produce valid signatures on Bitcoin transactions. 

We handle the complex x-only negation conditions of BIP340 by introducing new methods on the `Ciphersuite` trait, which are generally identity functions in the case of other ciphersuites, but which in the taproot suite will conditionally negate certain values at critical times to maintain compatibility with BIP340's x-only public keys. These methods are usually denoted `C::effective_xyz(...)`.

To support BIP341 taptree tweaking, we add two new types, `SigningTarget` and `SigningParameters`, and modify the signature package to accept any type that transforms `Into` `SigningTarget`. For backwards compatibility, this includes any type which implements `AsRef<[u8]>`. These convert into `SigningTarget` paired with `SigningParameters::default()`. The `SigningParameters` can be modified to commit signatures to a specific tapscript merkle root tweak, allowing FROST groups to modify (tweak) their group keys with taproot script-spending paths, while retaining the use of the key-spending path.

Some additional features we can consider adding:
- [ ] Adaptor signature support
- [ ] BIP32 key derivation
- [ ] Higher-level (nested) multisig support

i have a separate feature branch, [`taproot-bip32`](https://github.com/conduition/frost/compare/add-secp256k1-tr..taproot-bip32), based off of this branch which adds BIP32 support to the `frost-secp256k1-tr` crate. See [this discussion](https://github.com/ZcashFoundation/frost/discussions/636) for more background. It would be nice to be able to PR that branch into this repo as well, so that FROST keys can be used in hierarchical and descriptor wallets. For now I am leaving that separate to reduce conflation of features, but they can be merged together if desired.


conduition:
This PR clearly needs to be resynced with `main`, I will have a go at that soon after conferring with my clients

conduition:
As noted by @jesseposner during the round table, the raw DKG output is not safe to use in a BIP341 context. I'm going to update this PR to reflect that, which might entail some API changes.

conradoplg:
> As noted by @jesseposner during the round table, the raw DKG output is not safe to use in a BIP341 context. I'm going to update this PR to reflect that, which might entail some API changes.

I had the impression that this PR would adjust the sign of the X coordinate during signing? That would solve the issue.

(I ask this because I remember the first thing I wanted to do in this PR was to make that during key generation only, but that's not possible by @jesseposner's point)

jesseposner:
There's a negation algorithm in the FROST signing BIP: https://github.com/siv2r/bip-frost-signing

It gets a little tricky to account for both BIP32 tweaks and taproot script tweaks (TapTweaks), which is why the algorithm is a bit complex. The algorithm is implemented here: https://github.com/BlockstreamResearch/secp256k1-zkp/pull/278

It's also the same algorithm used in the MuSig2 BIP and implementation.

conduition:
@conradoplg Yes, this PR does handle key negation properly at signing time (i haven't read Jesse's linked negation algorithm yet), but how would key negation at signing time prevent a cosigner from injecting a malicious tweak into the group key at DKG time?

Maybe it'd help to actually write out the steps the attacker will do, so we're sure we're all talking about the same thing.

Let's say we run a DKG to construct a simple 2-of-2 FROST group key. Honest participant $1$ samples their random coefficients $\\{a_{10}, a_{11} \\}$ and broadcasts their commitment $C_1 = \\{ \phi_{10}, \phi_{11} \\} = \\{ a_{10} G, a_{11} G \\}$. 

A malicious participant $2$ also samples random coefficients $\\{ a_{20}', a_{21} \\}$, but now they can inject a malicious tweak as follows.

1. Compute the 'internal' group pubkey $Y'$.

$$ Y' = \phi_{10} + a_{20}' \cdot G $$

2. Given some malicious tapscript tree merkle root $m$, compute the tweaked group public key $Y$:

$$ t = H(Y'\ \|\|\ m) $$

$$ Y = Y' + t G $$

3. Tweak the constant term coefficient by adding $t$.

$$ a_{20} = $a_{20}' + t $$

4. Compute the DKG commitment using the maliciously tweaked coefficient. 
 
$$ C_2 = \\{\phi_{20}, \phi_{21}\\} =  \\{a_{20} G, a_{21} G\\} $$ 

5. Continue the FROST DKG protocol as normal.
 

The resulting group pubkey accepted by the group will be:

$$ 
\begin{align}
Y &= \phi_{10} + \phi_{20} \\\\
  &= \phi_{10} + a_{20} \cdot G \\\\
  &= \phi_{10} + (a_{20}' + t) \cdot G \\\\ 
  &= \phi_{10} + a_{20}' \cdot G  + t G \\\\
  &= Y' + H(Y'\ \|\|\ m) \cdot G \\\\
\end{align}
$$

The DKG will succeed because the malicious participant does actually know the coefficients of their secret contribution polynomial and can produce a valid proof of possession. The only drawback of this attack is that it depends on the attacker waiting to hear everyone else's DKG commitments first, in order to compute the rogue tweak. And naturally it doesn't work if the group intends to use BIP32 or BIP341 tweaks after DKG, but if the group uses the raw DKG output pubkey $Y$ in a P2TR script on chain, then the attacker can use the script-spending path described by $m$ to steal the coins.

MatthewLM:
@conduition Perhaps a simple solution is to tweak the DKG output in a way similar to BIP32? It could be as simple as tweaking by the group public key hash? Then the private shares, public shares and group key provided by the library can include this tweak with no API changes needed.

conduition:
@MatthewLM I like that approach, it seems simple and elegantly prevents the footgun, without limiting how participants use their group key. The group can still add further tweaks (e.g. BIP341, BIP32, etc) at signing time too. I spent last night trying to break it, but couldn't see anything wrong.

@jesseposner @conradoplg What do you guys think?

conduition:
I'm starting the process of resyncing this PR with the upstream v2.0.0 changes on `main`.

jesseposner:
This sounds like what @real-or-random suggested here: https://github.com/BlockstreamResearch/bip-frost-dkg/issues/41#issuecomment-2383786511

MatthewLM:
Isn't the key already tweaked at sign-time anyway? Looking at my own code, I've used an earlier version of this branch and I see that a TapTweak is already done. I was previously able to create successful Taproot signatures with key-spend tweaks. I'm not sure any changes are needed.

conduition:
Resyncing is done but not yet pushed to this branch. You can preview the changes here: https://github.com/ZcashFoundation/frost/compare/main...conduition:frost:add-secp256k1-tr-reloaded

@conradoplg Do you have any preference for how i merge the updates into this branch? If it's not important, i'll just force-push the new commit history over this PR. I tagged @mimoo and zebra-lucky as co-contributors in the commit messages but i'm not sure how else to properly attribute.

> Isn't the key already tweaked at sign-time anyway?

@MatthewLM Not always. The current code will, by default, sign without any tweak. The untweaked group pubkey which those signatures would verify under is susceptible to rogue tweak attacks at DKG-time. 

I think a static unspendable tweak of the group key at DKG time is the ideal scenario here, as this ensures that no sane caller will ever accidentally pay money to a group pubkey which could contain a rogue tweak. I'll see about adding this soon after merging the re-synced branch into this one.

conduition:
Just to formalize this 'tweak the DKG output' idea a little...

Every FROST participant receives three things from the DKG:

- A signing share $s_i = \sum_{j=1}^n f_j(i)$
- A set of $n$ verification shares $\\{Y_1 ... Y_n\\}$. Each share is computed using the commitment coefficients: $Y_i = \sum_{j=1}^n \sum_{k=0}^{t-1} i^{k} \cdot \phi_{jk}$
- A group verification key $Y = \sum_{i=1}^n \phi_{i0}$

To prevent malicious tweaks (inserted into some $\phi_{i0}$), we compute a tweak $t = H_{\text{TapTweak}}(Y)$. We then additively tweak all three DKG outputs with $t$.

- $Y_i' = Y_i + tG$
- $Y' = Y + tG$
- $s_i' = s_i + t$

The participants can do this implicitly after DKG, without any extra communication rounds. The shares are still valid and the resulting group key $Y'$ is certain not to have any tapscript spending path.

conradoplg:
> @conradoplg Do you have any preference for how i merge the updates into this branch? If it's not important, i'll just force-push the new commit history over this PR. I tagged @mimoo and zebra-lucky as co-contributors in the commit messages but i'm not sure how else to properly attribute.
> 

It doesn't matter, PRs are squashed in this repo anyway. GitHub will include the contributors in the commit message when it merges.

Let me know if you need help solving conflicts.

frost-secp256k1-tr/src/lib.rs
```diff
@@ -0,0 +1,755 @@
...
+    /// However, for FROST, this is not strictly required as the group key cannot be
+    /// poisoned as long as the DKG procedure is conducted correctly.
+    /// Thus, the [`Default`] trait implementation of taproot `SigningParameters`
+    /// sets `tapscript_merkle_root` to `None`.
+    ///
+    /// If 3rd party observers outside the FROST group must be able to verify there
+    /// is no hidden script-spending path embedded in the FROST group's taproot output key,
+    /// then you should set `tapscript_merkle_root` to `Some(vec![])`, which proves
+    /// the tapscript commitment for the tweaked output key is unspendable.
+    pub tapscript_merkle_root: Option<Vec<u8>>,
```
MatthewLM, line 262: @conduition It was a while ago that I wrote my own code but I went back and I see this is where the tweak can be provided if this is set to an empty vector. If this is set to None for a key-spend then that is where the problem lies.

In my own code I set this to None when providing signatures for a script-path spend. I don't believe there is any issue with that. I pass an empty vector when a key-spend is done without a MAST hash which then includes the necessary tweak.

Provided that consumers of the library are passing an empty vector when doing a MAST-free key-spend, I don't see there being an issue. However there might be a problem if this is used incorrectly and None is accidentally used. An additional tweak as part of the DKG process might still be a good idea to protect against misuse.

Maybe the API can be improved with factory methods such as `SigningParameters::keySpend` and `SigningParameters::scriptSpend` so it's clear when a key is used for key-path or script-path signatures and where key-path keys are always tweaked.

I've written a Dart wrapper that uses these high-level Dart constructors:

```dart
  /// Used for key-spend signatures. This will always tweak the internal key. If
  /// there is no MAST root hash provided, the tweak will be done without a
  /// MAST hash for Taproot programs that only have a key-path.
  SignDetails.keySpend({
    required Uint8List message,
    Uint8List? mastHash,
  }) : this(message: message, mastHash: mastHash ?? Uint8List(0));

  /// Used for script-spend signatures. This does not tweak the key allowing for
  /// direct signing with a key specified in a Tapscript.
  SignDetails.scriptSpend({ required Uint8List message })
    : this(message: message, mastHash: null);

```

conduition: > In my own code I set this to None when providing signatures for a script-path spend. I don't believe there is any issue with that.

Yes, you're correct. In fact I have some clients who are doing exactly this (script-path spending with FROST signatures in the script witness)

> Provided that consumers of the library are passing an empty vector when doing a MAST-free key-spend, I don't see there being an issue.

That's exactly the issue: We want security _by default_. Key-spending is the default for P2TR, so we can't allow key-spending users of this crate to footgun themselves just because they didn't know to explicitly configure their verifying key with the right arcane parameter of `Some(vec![])`.

> Maybe the API can be improved with factory methods such as SigningParameters::keySpend and SigningParameters::scriptSpend so it's clear when a key is used for key-path or script-path signatures and where key-path keys are always tweaked.

Technically you can use tweaked or untweaked BIP340 signatures for either key-spending or script-spending, there's no hard relationship there. 

- Maybe I'd want to tweak a pubkey inside the MAST tapscript leaf to commit to some custom data. Even if bitcoin nodes don't care or validate that tweak, it's still a useful option for me in off-chain protocols.
- Likewise, it's perfectly fine for a single-signer to use an untweaked plain public key as a P2TR output and then key-spend from that.

The only dangerous combination is using an untweaked _linearly aggregated_ pubkey in a P2TR output script, _if_ the keys which compose that aggregated pubkey come from untrusted sources.

MatthewLM: Yes, I think it's fair to tweak the DKG output to avoid misuse. I included named constructors in my code to avoid misuse but it's not difficult to add a tweak during the DKG process.

conduition:
I've resynchronized this branch with upstream `main` to be compatible with the v2.0.0 codebase. Tests are passing and clippy is happy. I'll add DKG tweaking soon in a new commit

0xBEEFCAF3 (approved):
tACK  802de7a5c09faba9e5f3b7347bbed1b52a8d1dff
FWIW I ran some functionality tests against a regtest bitcoind setup
My test suite can be found [here](https://github.com/0xBEEFCAF3/frost-playground/)
Tested:
✅ Key Spend with Trusted Dealer Setup
✅ Script Spend with Trusted Dealer Setup
✅ Key Spend with DKG Setup
✅ Script Spend with DKG Setup

conradoplg:
Just FYI I'm actively reviewing this. It looks good, but I think it might be possible to simplify some of the code so I'm working on that.

conradoplg:
I pushed a cleanup of the code.

My main motivation was to minimize changes to `frost_core` and also try not to include BIP-340-specific code in it (e.g. `y_is_odd()`, negation methods etc.). This is because `frost_core` was audited and I want to minimize the risk of introducing bugs in it.

I change the approach of the additional `Ciphersuite` methods to be more "high-level".

One aspect which I'm not sure about is tweak handling. In my commit I reverted the `SigningTarget/Parameters` types because I did not want to change the encoding of the `SigningPackage`. I feel like additional information that needs to be communicated should be handled by the caller.

This means that, with my changes, tweak handling is not longer "transparent". You will need to call e.g. `frost_secp256_k1_tr::round2::sign_with_tweak(signing_package, signing_nonces, key_package, merkle_root)` and you won't be able to use just `frost_core` (e.g. `frost_core::round2::sign<frost_secp256_k1_tr::Secp256K1Sha256TR>::sign(...)` if you want to use tweaks. Is that acceptable for users of this ciphersuite?

There is also some stuff missing:

- DKG test vectors are not passing. I changed the proof-of-knowledge handling to _not_ use "even Y" because I don't think there is a need to. But this means that the test vectors need to be updated. I can eventually manage to fix those, but if whoever created them could update them, I would appreciate that a lot.
- I need to fix conflicts with `main`
- Some documentation improvements (probably its README should have information about tweak handling, so I'm waiting on a decision on that API)
- I thinking about using a feature that needs to enabled in frost_core to activate these additional trait methods, so that we don't have to do a major version bump for this.
- Is there an easy way to test if tweak handling is correct? i.e. if a tx signed with it would be really accepted. (Or @0xBEEFCAF3 would you mind rerunning your tests? Do they cover the usage with tweaks?)

I apologize for changing everything so much, but I feel like this approach has some advantages and makes everything a bit easier to reason about. I really appreciate all the effort that went into this.

You can look at the individual commit to see the changes from the original approach, and the overall PR diff to see differences with `main` (which should be a bit smaller now, in particular regarding `frost_core`).

Looking forward to your feedback.

conduition:
Thanks @conradoplg, I'll try to take a look at these changes soon.

conradoplg:
Thinking about it I don't think we can get away with not using even-Y for DKG proof-of-knowledge since the signature serialization only encodes X and the PoK is represented as a `Signature`. The tests are not catching this likely because they don't do serialization roundtrips (that's #688). I'll look into it.

conradoplg:
> Thinking about it I don't think we can get away with not using even-Y for DKG proof-of-knowledge since the signature serialization only encodes X and the PoK is represented as a `Signature`. The tests are not catching this likely because they don't do serialization roundtrips (that's #688). I'll look into it.

Ah nevermind. I _am_ using even-Y for the R component of the signature. It's just the DKG `round1::Package` public key which is not using even-Y so adjusting the test vectors might be enough.

frost-secp256k1-tr/src/lib.rs
```diff
@@ -327,68 +304,176 @@ impl Ciphersuite for Secp256K1Sha256 {
...
+        ctx.is_group_commitment_even =
+            (!group_commitment.clone().to_element().to_affine().y_is_odd()).into();
+        let group_commitment = if !ctx.is_group_commitment_even {
+            GroupCommitment::<S>::from_element(-group_commitment.to_element())
+        } else {
+            group_commitment
+        };
+
+        Ok(group_commitment)
+    }
```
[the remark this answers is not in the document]
conradoplg: This sounds great! I'd like to avoid having to pass the group commitment around if we could, but this feels much better than having to add `Context`. Please go ahead if you want, thanks.

frost-core/src/tests/ciphersuite_generic.rs
```diff
@@ -196,8 +194,7 @@ pub fn check_sign<C: Ciphersuite + PartialEq, R: RngCore + CryptoRng>(
     key_packages: BTreeMap<frost::Identifier<C>, frost::keys::KeyPackage<C>>,
     mut rng: R,
     pubkey_package: PublicKeyPackage<C>,
-    signing_target: SigningTarget<C>,
-) -> Result<(SigningTarget<C>, Signature<C>, VerifyingKey<C>), Error<C>> {
```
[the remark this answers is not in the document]
conradoplg: I added a test in `tweaking_tests.rs`, where I basically copied the README test and added tweaking. I don't like the repetition but I think it will do it for now.

frost-secp256k1-tr/src/lib.rs
```diff
@@ -421,174 +506,27 @@ impl Ciphersuite for Secp256K1Sha256 {
...
+        // From BIP-341:
+        // > If the spending conditions do not require a script path, the output
+        // > key should commit to an unspendable script path instead of having
+        // > no script path. This can be achieved by computing the output key
+        // > point as Q = P + int(hashTapTweak(bytes(P)))G.
+        let merkle_root = key_package.verifying_key().to_element().to_affine().x();
+        Ok((
+            key_package.tweak(Some(merkle_root)),
+            public_key_package.tweak(Some(merkle_root)),
+        ))
```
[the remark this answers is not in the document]
conradoplg: Oops, good catch. Should've double checked that. Feel free to fix it if you're changing the code, otherwise I can fix it later.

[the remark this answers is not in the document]
conduition: Fixed in f9237f5bc2f58b385b56f10164103bdcf4890a4d :smile: Also fixes a related bug where key-package tweaking caused a bout of conditional key negation

conduition:
@conradoplg Tests are passing for me on this branch :tada: 

Now that we've got that sorted out, i'm going to do [the pruning i suggested here](https://github.com/ZcashFoundation/frost/pull/730#discussion_r1817410480) to hopefully reduce code surface area. Fingers crossed it works. Might take a few days but i'll check in again soon :slightly_smiling_face:

0xBEEFCAF3:
>Is there an easy way to test if tweak handling is correct? i.e. if a tx signed with it would be really accepted. (Or @0xBEEFCAF3 would you mind rerunning your tests? Do they cover the usage with tweaks?)

@conradoplg yes happy to run [the testsuite](https://github.com/0xBEEFCAF3/frost-playground/) again. I added a test to cover the taptweak case. In this test, we spend using the FROST key through the default key spend path, while also including optional tapleaf script paths.

All tests are passing against the latest commit ([f9237f5](https://github.com/ZcashFoundation/frost/pull/730/commits/f9237f5bc2f58b385b56f10164103bdcf4890a4d))

frost-secp256k1-tr/src/lib.rs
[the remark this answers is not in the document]
conduition: @conradoplg I've pushed these changes, `Context` is no more! :rocket:

frost-secp256k1-tr/src/lib.rs
```diff
@@ -758,12 +758,10 @@ pub mod keys {
         fn tweak<T: AsRef<[u8]>>(self, merkle_root: Option<T>) -> Self {
             let t = tweak(&self.verifying_key().to_element(), merkle_root);
             let tp = ProjectivePoint::GENERATOR * t;
-            let public_key_package = self.into_even_y(None);
-            let verifying_key =
-                VerifyingKey::new(public_key_package.verifying_key().to_element() + tp);
+            let verifying_key = VerifyingKey::new(self.verifying_key().to_element() + tp);
```
conradoplg: I thought this was needed? BIP-341 uses

```
P = lift_x(int_from_bytes(pubkey))
Q = point_add(P, point_mul(G, t))
```

where `lift_x` always returns an even-Y point. Am I missing something?

conradoplg:
> Now that we've got that sorted out, i'm going to do [the pruning i suggested here](https://github.com/ZcashFoundation/frost/pull/730#discussion_r1817410480) to hopefully reduce code surface area. Fingers crossed it works. Might take a few days but i'll check in again soon 🙂

Thanks, it looks great. I'll sort out the conflicts next.

frost-secp256k1-tr/src/lib.rs
conduition: I think you're right, this change might have been a mistake. It wouldn't have shown in the signing tests, because the FROST signatures are still valid under the output pubkey, but it would mean the taproot spending path would be unusable, because the "internal key" (`pubkey` in your cited example) doesn't result in the same point X-coordinate as the one put on-chain.

I'll add a test to confirm this is this the case, and then fix it if so.

conduition: Fixed in 5dc16a29df7849b93e7184b9fb09c56e9177406a, thanks for spotting this :)

conduition: Fixed in 5dc16a29df7849b93e7184b9fb09c56e9177406a - Thanks for spotting this :)

conradoplg:
Everything looks great, thank you all for the contributions. I'll merge this soon unless someone wants to add anything?

conradoplg (approved):

conduition:
Thank you very much @conradoplg :heart:

real-or-random:
Hi, sorry for being late to the party. I'm just curious: Is there any specific reason why someone in the Zcash ecosystem would need this (e.g., swaps with Bitcoin)? Or are you "just" making this available to the world because you feel it's useful that there is a Rust implementation for experimentation and given that Bitcoin is widespread? In either case, it's nice to see a second independent implementation! :)

conradoplg:
> Hi, sorry for being late to the party. I'm just curious: Is there any specific reason why someone in the Zcash ecosystem would need this (e.g., swaps with Bitcoin)? Or are you "just" making this available to the world because you feel it's useful that there is a Rust implementation for experimentation and given that Bitcoin is widespread? In either case, it's nice to see a second independent implementation! :)

Our original rationale is that BIP-340 is probably the biggest application of FROST and it made sense to support that; this repo was never meant to be Zcash-specific (the Zcash ciphersuites are not even here, but in https://github.com/ZcashFoundation/reddsa/ , though I'm thinking of moving them here too...).

But now it _does_ have an application for Zcash; the upcoming [Zcash Shielded Assets](https://zips.z.cash/zip-0227) will use BIP-340 signatures for issuance, and that's also a great use case for FROST.

PR #741, opened 2024-10-01
Add missing `derive(Getters)` for `dkg::{round1, round2}`

### LINKED ISSUE #737: Derive `Getters` for `frost_core::keys::dkg::round1::SecretPackage`
For example, I have a public place with `round1_packages` (in the blockchain), and I want to do `round1_packages.remove(&self.round1_secret_package.identifier())` to remove the participant identifier from `round1_packages` and pass it to `dkg::part2()`.

StackOverflowExcept1on:
Resolves #737

### DIFF
--- a/frost-core/CHANGELOG.md
+++ b/frost-core/CHANGELOG.md
@@ -10,6 +10,7 @@ Entries are listed in reverse chronological order.
   but it's likely to not require any code changes since most ciphersuite
   implementations are probably just empty structs. The bound makes it possible
   to use `frost_core::Error<C>` in `Box<dyn std::error::Error>`.
+* Added getters to `round1::SecretPackage` and `round2::SecretPackage`.
 
 ## 2.0.0-rc.0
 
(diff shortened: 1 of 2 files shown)


StackOverflowExcept1on:
@conradoplg could you take a look? real use cases:
1. smart contract has in its storage `round1_packages: BTreeMap<Identifier<C>, round1::Package<C>>` and you as a participant want to remove your `identifier` from `round1_packages`: https://github.com/StackOverflowExcept1on/roast/blob/c31716b0eeefcaf6ab7029e7bb882334133c07c2/roast-core/src/dkg.rs#L311
2. participant wants to execute `dkg::part3()`, but with a small difference - bad participants will be recorded in `round2_culprits_set`, and then passed to smart contract, so that it will apply sanctions to them (for example, take away deposit of bad DKG participants): https://github.com/StackOverflowExcept1on/roast/blob/c31716b0eeefcaf6ab7029e7bb882334133c07c2/roast-core/src/dkg.rs#L362

StackOverflowExcept1on:
I skipped the coefficients field because there is already a method that returns `&[T]` instead of `&Vec<T>`, which is a more correct type (internal method). Unfortunately, in derive getters you can't choose a more correct type (e.g. u16 instead of &16).

conradoplg:
Ah of course, thanks

conradoplg (approved):

PR #757, opened 2024-10-16
Expose to_element() for GroupCommitment

natalieesk:


frost-core/src/lib.rs
```diff
@@ -478,7 +478,8 @@ where
     C: Ciphersuite,
 {
     /// Return the underlying element.
-    #[cfg(feature = "internals")]
+    #[cfg_attr(feature = "internals", visibility::make(pub))]
+    #[cfg_attr(docsrs, doc(cfg(feature = "internals")))]
     pub fn to_element(self) -> <C::Group as Group>::Element {
```
conradoplg, line 483: `#[cfg_attr(feature = "internals", visibility::make(pub))]` makes it `pub` when `internals` is enabled, but it is already `pub`, so we should make it `pub(crate)`. (Sorry, I just realized this now)

```suggestion
    pub(crate) fn to_element(self) -> <C::Group as Group>::Element {
```

conradoplg (changes requested):

frost-core/src/lib.rs
natalieesk: Ah! Thank you. Ok

conradoplg (approved):

PR #764, opened 2024-11-04
fix(changelog): now it's possible to identify culprit in part3(), but only in unreleased version

StackOverflowExcept1on:
@conradoplg


conradoplg:
Thanks. Yep, sorry about that, it will make the next release which shouldn't take long. Let us know if that's a blocker for you.

PR #766, opened 2024-11-08
Refresh Shares with DKG

### LINKED ISSUE #633: Make server ciphersuite-blind
Problem: the server needs to know the ciphersuite and only works with one. This is because the Args/Output structs have FROST structs in them, which are (de)serialized as needed. This requires specifying a ciphersuite.

Solution: the server doesn't need to do the serialization roundtrip. It can receive FROST structs serialized and just pass them around.

Coding approach suggestion:

- Change the FROST structs in `server/src/types.go` to use `Vec<u8>` instead, e.g. `pub commitments: Vec<frost::round1::SigningCommitments>` becomes `pub commitments: Vec<Vec<u8>>`.
  - By default serde encodes these `Vec<u8>` as integer arrays which is awkward. It's possible to encode them as hex strings (in human-readable formats like JSON) or directly binary. Check how we do that in `SigningFormat::message` field in `frost-core` for a reference on how to do that with `serde(serialize_with)`.
  - For type safety it's possible to use newtypes for each struct, e.g. define a `EncodedSigningCommitments(Vec<u8>)` and use that instead.

### LINKED ISSUE #663: Feature: Remove signers or refresh shares with DKG
## Motivation
There may be situations where you may want to _remove_ signers. 

## Design
To do this you would need to perform a resharing protocol and leave out the relevant participant.

This is done by basically redoing the DKG with a "zero" secret, and adding the new shares to the old ones, while keeping only the participants that are wanted

## References
<add here>

natalieesk:
This closes #663 

This adds the refresh share functionality with DKG. It needs some tidying up but that will be done in another PR.

### DIFF
--- a/frost-core/src/keys/refresh.rs
+++ b/frost-core/src/keys/refresh.rs
@@ -8,14 +8,18 @@ use alloc::collections::BTreeMap;
 use alloc::vec::Vec;
 
 use crate::{
+    keys::dkg::{compute_proof_of_knowledge, round1, round2},
     keys::{
-        generate_coefficients, generate_secret_shares, validate_num_of_signers,
-        CoefficientCommitment, PublicKeyPackage, SigningKey, SigningShare, VerifyingShare,
+        evaluate_polynomial, generate_coefficients, generate_secret_polynomial,
+        generate_secret_shares, validate_num_of_signers, CoefficientCommitment, PublicKeyPackage,
+        SigningKey, SigningShare, VerifyingShare,
     },
-    Ciphersuite, CryptoRng, Error, Field, Group, Identifier, RngCore,
+    Ciphersuite, CryptoRng, Error, Field, Group, Header, Identifier, RngCore,
 };
 
-use super::{KeyPackage, SecretShare, VerifiableSecretSharingCommitment};
+use core::iter;
+
+use super::{dkg::round1::Package, KeyPackage, SecretShare, VerifiableSecretSharingCommitment};
 
 /// Generates new zero key shares and a public key package using a trusted
 /// dealer Building a new public key package is done by taking the verifying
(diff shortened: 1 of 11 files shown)


frost-core/src/keys/refresh.rs
```diff
@@ -114,3 +118,260 @@ pub fn refresh_share<C: Ciphersuite>(
 
     Ok(new_key_package)
 }
+
+/// Part 1 of refresh share with DKG. A refreshing_key is generated and a new package and secret_package are generated.
+/// The identity commitment is removed from the packages.
```
conradoplg, line 123: Nitpick: This is an implementation detail, I don't think it needs to be part of the public documentation

frost-core/src/keys/refresh.rs
```diff
@@ -114,3 +118,260 @@ pub fn refresh_share<C: Ciphersuite>(
...
+    let (coefficients, commitment) =
+        generate_secret_polynomial(&refreshing_key, max_signers, min_signers, coefficients)?;
+
+    // Remove identity element from coefficients
+    let mut coeff_comms = commitment.0;
+    coeff_comms.remove(0);
+    let commitment = VerifiableSecretSharingCommitment::new(coeff_comms.clone());
+
+    let proof_of_knowledge =
+        compute_proof_of_knowledge(identifier, &coefficients, &commitment, &mut rng)?;
```
conradoplg, line 149: This builds a kind of "nonsensical" proof of knowledge since it uses the first coefficient (zero) and the commitment to the _second_ coefficient (now first in the list). Which is not an issue, since we simply don't verify it. But it may be confusing to whoever is reading the source.

My suggestion is to directly build a dummy `Signature` using `Signature::new()` and passing a generator element (`generator()`) and a 1 scalar (`one()`), and adding a comment saying that we are building a dummy proof of knowledge because the secret is always zero and the proof isn't needed (and couldn't be encoded since it would have an identity element)

conradoplg (approved):
Looks great, thanks. I added a couple of comments, feel free to mark them as solved and address them in the refactoring PR if you prefer to.

PR #767, opened 2024-11-10
chore(frost-p256, frost-secp256k1): do not use allocator for domain separators

StackOverflowExcept1on:
@conradoplg


conradoplg (approved):
Neat, thanks!

Issue #769, opened 2024-11-10
Switch to fork of ed448


StackOverflowExcept1on:
There is fork that also has no_std support, fixed warnings, etc: https://github.com/mikelodder7/Ed448-Goldilocks

conradoplg:
I understand the desire but we'd like to avoid pointing to less popular forks (and the original crate was already not too mainstream). Ideally https://github.com/crate-crypto/Ed448-Goldilocks/issues/35 would be the best for us.

I'll leave this open while we consider the options, thanks!

StackOverflowExcept1on:
I don't know what needs to happen for https://github.com/crate-crypto/Ed448-Goldilocks to become part of RustCrypto... They have a lot of warnings when building project, some things are not fully implemented. Even if it becomes RustCrypto, the API will probably be different.

StackOverflowExcept1on:
@conradoplg Technically `ed448-goldilocks-plus 0.16.0` is same as `RustCrypto/elliptic-curves/ed448-goldilocks`: https://github.com/RustCrypto/elliptic-curves/blob/bc0960f8b4259026a96ae3c85c7e420d9f6d8a8b//README.md?plain=1#L87. Also `ed448-goldilocks-plus 0.16.0` from crates.io requires `rand_core 0.6` rather than `0.9` and we could use it to make FROST `#![no_std]` library without waiting for all the `RustCrypto` crates on new version of `rand_core` and with the new MSRV policies.

PR #772, opened 2024-11-11
feat: bump MSRV to 1.81, make all crates no_std

### LINKED ISSUE #768: Bump MSRV to 1.81
Motivation:
1. We can get rid of `lazy_static!` in favor of [`std::sync::LazyLock`](https://doc.rust-lang.org/std/sync/struct.LazyLock.html)
2. We can use [`{integer}::div_ceil`](https://doc.rust-lang.org/std/primitive.u32.html#method.div_ceil)
    https://github.com/ZcashFoundation/frost/blob/ca33dad959e55bc66d5feac53a1275faae79377b/frost-core/src/scalar_mul.rs#L28
3. Crates from RustCrypto already have pre-releases that require Rust 1.73+ or even Rust 1.81+ in some cases. For example: https://github.com/RustCrypto/elliptic-curves/blob/master/k256/Cargo.toml
4. Starting with 1.81 we can use [`core::error::Error`](https://doc.rust-lang.org/core/error/trait.Error.html) and remove the `thiserror-nostd-notrait` dependency in favor of `thiserror`: https://github.com/dtolnay/thiserror/pull/373

StackOverflowExcept1on:
Resolves #768
Should be merged after #769

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -66,10 +66,9 @@ jobs:
   build_no_std:
     name: build with no_std
     runs-on: ubuntu-latest
-    # Skip ed448 which does not support it.
     strategy:
       matrix:
-        crate: [ristretto255, ed25519, p256, secp256k1, rerandomized]
+        crate: [ed448, ristretto255, ed25519, p256, secp256k1, rerandomized]
     steps:
     - uses: actions/checkout@v4.1.7
     - uses: dtolnay/rust-toolchain@master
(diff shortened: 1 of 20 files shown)


StackOverflowExcept1on:
@conradoplg

conradoplg:
Thank you. We'll close this for now because we don't want to bump to 1.81 yet, and we want to wait a bit to see how that Ed448 forks settles. We'll revisit this PR later.

PR #774, opened 2024-11-13
fix: typos in documentation files

leopardracer:
This pull request contains changes to improve clarity, correctness and structure.

**Description correction:**
Corrected "needs an unique" to "needs a unique".
Corrected "If there is not" to "If there is no".

Please review the changes and let me know if any additional changes are needed.

PR #782, opened 2024-11-19
secp256k1-tr: fix no-std support

conradoplg:
Also regenerates some .md files that were missing in #774

### DIFF
--- a/frost-secp256k1-tr/src/lib.rs
+++ b/frost-secp256k1-tr/src/lib.rs
@@ -1,3 +1,4 @@
+#![cfg_attr(not(feature = "std"), no_std)]
 #![allow(non_snake_case)]
 #![deny(missing_docs)]
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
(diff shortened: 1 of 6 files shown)


frost-secp256k1-tr/src/lib.rs
```diff
@@ -10,6 +11,7 @@ extern crate alloc;
 use alloc::borrow::Cow;
 use alloc::borrow::ToOwned;
 use alloc::collections::BTreeMap;
+use alloc::vec;
 use alloc::vec::Vec;
```
StackOverflowExcept1on, line 15:
```suggestion
use alloc::vec::{self, Vec};
```

StackOverflowExcept1on: same for `use alloc::{Cow, Borrow}`

StackOverflowExcept1on:
We could also apply the recent changes #767 to frost-secp256k1-tr.

conradoplg:
> We could also apply the recent changes #767 to frost-secp256k1-tr.

Good call, added in b558f41d

natalieesk (approved):
Looks good

PR #791, opened 2024-11-25
Explanation of the Importance of Correcting These Errors

futreall:
1. **"transaprent" → "transparent"**
   - This typo disrupts the proper spelling of the word and can confuse the reader. In technical documentation, accuracy is critical, and an error in a key term like "transparent" could lead to misunderstandings.

2. **"they to go through" → "them to go through"**
   - This is a grammatical error. In English, after the verb "allow," the object pronoun "them" should be used (instead of "they"). Using "they" in this context makes the sentence grammatically incorrect and harder to understand.

3. **"communication to each other" → "communicate with each other"**
   - This is a mistake in using the noun "communication" instead of the verb "communicate." In this context, it's more appropriate to use the verb "communicate," as it expresses an action rather than an abstract noun. This is important for proper sentence construction.

5. **"catastrophical" → "catastrophic"**
   - "Catastrophical" is not a valid word in English. The correct term is "catastrophic." An error in such an important term could cause confusion and lower the professionalism of the documentation.

### DIFF
--- a/book/src/zcash/technical-details.md
+++ b/book/src/zcash/technical-details.md
@@ -120,4 +120,5 @@ instead of publishing them on the blockchain.
 ```info
 Another idea is to extend lightwalletd servers to support messaging, since
 wallets are all already connected to a server (not the same one, so inter-server
-communications would be also needed) ```
+communications would be also needed)
+```


book/src/zcash/technical-details.md
```diff
@@ -121,4 +121,4 @@ instead of publishing them on the blockchain.
 Another idea is to extend lightwalletd servers to support messaging, since
 wallets are all already connected to a server (not the same one, so inter-server
 communications would be also needed)
-```
\ No newline at end of file
+```
```
mpguerra, line 124: Can you revert this line change? I don't see why it was removed and then added back in. Thank you!

mpguerra (changes requested):
Thank you for correcting those typos!

PR #806, opened 2024-12-11
fix(frost-secp256k1-tr): add missing `pub mod refresh` and missing test

StackOverflowExcept1on:
@conradoplg missed in #766
also fixes CI: `cargo run --bin gencode -- --check`

I'm sure how to fix warnings:
```
WARNING: documentation for aggregate_with_tweak is not available in base file. This can mean it's a specific type for the ciphersuite, or that there is a bug in gencode
WARNING: documentation for sign_with_tweak is not available in base file. This can mean it's a specific type for the ciphersuite, or that there is a bug in gencode
WARNING: documentation for Tweak is not available in base file. This can mean it's a specific type for the ciphersuite, or that there is a bug in gencode
WARNING: documentation for EvenY is not available in base file. This can mean it's a specific type for the ciphersuite, or that there is a bug in gencode
```

### DIFF
--- a/frost-secp256k1-tr/src/lib.rs
+++ b/frost-secp256k1-tr/src/lib.rs
@@ -787,6 +787,7 @@ pub mod keys {
     }
 
     pub mod dkg;
+    pub mod refresh;
     pub mod repairable;
 }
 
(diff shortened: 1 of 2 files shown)


conradoplg (approved):
Thanks! The warnings are expected due to the ciphersuite-specific functions in the Taproot ciphersuite. I might eventually figure out a way for gencode to ignore those but it's fine for now.

PR #809, opened 2024-12-12
ci: fail on doc warnings

conradoplg:
This turns doc warnings into errors, making sure we don't miss them.


natalieesk (approved):
Looks good

PR #810, opened 2024-12-12
feat(frost-core): simplify trait bounds, don't use the allocator in some places

### LINKED ISSUE #807: `Copy + Clone` trait-bound can be simplified
because `Copy: Clone`
https://github.com/search?q=repo%3AZcashFoundation%2Ffrost%20%22Copy%20%2B%20Clone%22&type=code
https://doc.rust-lang.org/std/marker/trait.Copy.html

StackOverflowExcept1on:
Resolves #807
@conradoplg We discussed this earlier. This will also allow to encrypt `Serialization` for the 2nd round of DKG (without allocation). I haven't added `+ Default` yet as that would break the API, but I think even this implementation will be enough.


frost-core/src/traits.rs
```diff
@@ -37,7 +37,7 @@ pub trait Field: Copy + Clone {
         + Sub<Output = Self::Scalar>;
 
     /// A unique byte array buf of fixed length N.
-    type Serialization: AsRef<[u8]> + Debug + TryFrom<Vec<u8>>;
+    type Serialization: Copy + AsRef<[u8]> + AsMut<[u8]> + for<'a> TryFrom<&'a [u8]> + Debug;
```
conradoplg, line 40:
```suggestion
    type Serialization: AsRef<[u8]> + AsMut<[u8]> + for<'a> TryFrom<&'a [u8]> + Debug;
```

frost-core/src/traits.rs
```diff
@@ -102,7 +102,7 @@ pub trait Group: Copy + Clone + PartialEq {
     /// A unique byte array buf of fixed length N.
     ///
     /// Little-endian!
-    type Serialization: AsRef<[u8]> + Debug + TryFrom<Vec<u8>>;
+    type Serialization: Copy + AsRef<[u8]> + AsMut<[u8]> + for<'a> TryFrom<&'a [u8]> + Debug;
```
conradoplg, line 105:
```suggestion
    type Serialization: AsRef<[u8]> + AsMut<[u8]> + for<'a> TryFrom<&'a [u8]> + Debug;
```

frost-core/src/traits.rs
```diff
@@ -162,7 +162,11 @@ pub trait Ciphersuite: Copy + Clone + PartialEq + Debug + 'static {
 
     /// A unique byte array of fixed length that is the `Group::ElementSerialization` +
     /// `Group::ScalarSerialization`
-    type SignatureSerialization: AsRef<[u8]> + TryFrom<Vec<u8>>;
+    type SignatureSerialization: Copy
+        + AsRef<[u8]>
+        + AsMut<[u8]>
+        + for<'a> TryFrom<&'a [u8]>
+        + Debug;
```
conradoplg, line 169:
```suggestion
    type SignatureSerialization: AsRef<[u8]>
        + AsMut<[u8]>
        + for<'a> TryFrom<&'a [u8]>
        + Debug;
```

conradoplg (commented):
This looks good, but since it's a breaking change (in the SemVer sense, any Ciphersuite implementor will need to implement the additional traits) it will have to wait until we're ready to release 3.0.0.

I think the `Copy` bounds are not needed?

StackOverflowExcept1on:
> This looks good, but since it's a **breaking change**

@conradoplg I disagree, since implementations of `Ciphersuite` are more likely to just use `[u8; N]`. The `Copy` bundle is not used, but it's kind of an additional restriction (to have the user use `[u8; N]` for now, rather than wrapper type with `+ Default`).

StackOverflowExcept1on:
@conradoplg After some reseach about SemVer, I was only able to find 1 place where my PR breaks it: https://github.com/penumbra-zone/penumbra/blob/0e3d7966e3ff34bf74d7721a97848567f345e364/crates/crypto/decaf377-frost/src/traits.rs#L87. `type SignatureSerialization = Vec<u8>` and `type Serialization = Vec<u8>` are likely to cause an error, since both do not implement `Copy`. I found this through GitHub Search: https://github.com/search?q=%2F(%3F-i)impl+Ciphersuite+for%2F+NOT+is%3Afork&type=code. Should we be so worried and block it because of one breakdown?

frost-core/src/traits.rs
StackOverflowExcept1on: I also need `Copy` for my DKG implementation with encryption. Maybe we could reduce it to `Clone` instead of `Copy` (to also be compatible with types like `Vec<u8>`)?

PR #815, opened 2024-12-18
fix(frost-secp256k1-tr): empty merkle root tweak should still hash the x-only pk

0xBEEFCAF3:
I noticed that the addresses I was deriving with rust-bitcoin and the ones I created with the taptweak'd key using this library were not the same. The problem is in how the tweak is calculated in the absence of script paths.

Per [BIP-341](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki), if there are no script paths, the internal key should still be tapTweak'd by **tG, where t = TaggedHash(P_x)**. Before this commit, the internal key and the Taproot output key were the same if no script paths were used. This is because the tweak was the 0 scalar value, so Q = P + tG = P.

It is worth noting that Bitcoin's consensus rules will accept a non-taptweak'd internal key as it verifies a signature against whatever public key is used in the witness program. So the outputs are still spendable; however, they deviate from the BIP-341 spec. 

I also noticed that in [`post_dkg`](https://github.com/ZcashFoundation/frost/blob/main/frost-secp256k1-tr/src/lib.rs#L475), we are tweaking by `[0u8; 0]`. However, the comment above it correctly states the tweak should be `int(hashTapTweak(bytes(P)))G`. I can change this code slightly to be:
```rust
Ok((
    key_package.tweak::<&[u8]>(None),
    public_key_package.tweak::<&[u8]>(None),
))
```
This should just tweak with the hash of the x-only pk. Just lmk, happy to make this change as well.

### DIFF
--- a/frost-secp256k1-tr/src/lib.rs
+++ b/frost-secp256k1-tr/src/lib.rs
@@ -208,7 +208,11 @@ fn tweak<T: AsRef<[u8]>>(
     merkle_root: Option<T>,
 ) -> Scalar {
     match merkle_root {
-        None => Secp256K1ScalarField::zero(),
+        None => {
+            let mut hasher = tagged_hash("TapTweak");
+            hasher.update(public_key.to_affine().x());
+            hasher_to_scalar(hasher)
+        }
         Some(root) => {
             let mut hasher = tagged_hash("TapTweak");
             hasher.update(public_key.to_affine().x());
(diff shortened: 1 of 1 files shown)


conradoplg (changes requested):
Good catch, thanks. Looks good, but please include the change you mentioned to post_dkg if you can. Thanks!

0xBEEFCAF3:
> Good catch, thanks. Looks good, but please include the change you mentioned to post_dkg if you can. Thanks!

Done!

conradoplg (approved):

PR #835, opened 2025-01-14
all: fix issues related to features

conradoplg:
Some stuff is breaking in some combinations of features

### DIFF
--- a/frost-core/src/tests/refresh.rs
+++ b/frost-core/src/tests/refresh.rs
@@ -6,11 +6,10 @@ use crate::keys::generate_with_dealer;
 use crate::keys::refresh::{
     compute_refreshing_shares, refresh_dkg_part2, refresh_dkg_part_1, refresh_share,
 };
+#[cfg(feature = "serialization")]
+use crate::keys::{PublicKeyPackage, SecretShare};
 use crate::{self as frost};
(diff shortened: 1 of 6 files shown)


frost-core/src/tests/refresh.rs
```diff
@@ -6,11 +6,10 @@ use crate::keys::generate_with_dealer;
...
     compute_refreshing_shares, refresh_dkg_part2, refresh_dkg_part_1, refresh_share,
 };
+#[cfg(feature = "serialization")]
+use crate::keys::{PublicKeyPackage, SecretShare};
 use crate::{self as frost};
-use crate::{
-    keys::{KeyPackage, PublicKeyPackage, SecretShare},
-    Ciphersuite, Error, Identifier, Signature, VerifyingKey,
-};
+use crate::{keys::KeyPackage, Ciphersuite, Error, Identifier, Signature, VerifyingKey};
```
natalieesk, line 12: Really minor thing, but as you've refactored the crate::keys imports, could you also move keys::KeyPackage

natalieesk: Ah I see. Ok, thanks. No, that's fine.

natalieesk (approved):
Approved

PR #842, opened 2025-01-15
frost-core: fix visibility of coefficients()

conradoplg:
The new SecretPackage serialization code is breaking when the `internals` feature is disabled


natalieesk (approved):
Approved

PR #861, opened 2025-01-31
all: use OsRng instead of thread_rng()

conradoplg:
While our libraries don't do RNG directly and take an CryptoRng from the caller, we do have a bunch of tests and examples that do RNG and that people will inevitably copy.

Reading [`rand`s CHANGELOG](https://github.com/rust-random/rand/blob/c01aee7a138ff77657782069771bb11f120318d7/CHANGELOG.md) while reviewing #853, I noticed that it has a new `"rand is not a crypto library"` policy. Its `thread_rng` that we were using is also not fork-safe. Looking for alternatives I saw that librustzcash uses `OsRng`, which seems to be the right thing to do (even if it's still in `rand`, but it's basically a `getrandom()` wrapper that implements `CryptoRng`, which is exactly what we need).

This changes all usages of `thread_rng()` to use `OsRng` instead.

### DIFF
--- a/frost-ed25519/README.md
+++ b/frost-ed25519/README.md
@@ -11,10 +11,9 @@ scenario in a single thread and it abstracts away any communication between peer
 ```rust
 # // ANCHOR: tkg_gen
 use frost_ed25519 as frost;
-use rand::thread_rng;
 use std::collections::BTreeMap;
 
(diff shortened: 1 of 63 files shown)


natalieesk (approved):
Looks good

PR #869, opened 2025-02-08
fix duplicate

comfsrt:
the the → the


mpguerra:
We're no longer accepting these sorts of PRs

PR #875, opened 2025-02-16
Add tests for all test vectors in DKG

azuchi:
The current DKG test vector is only tested on participant 1's data. This PR applies the same tests to the data from participants 2 and 3.

Accordingly, the following modifications were made to the existing Test Vector:

### frost-ed448

Fix incorrect share generated from polynomials.

### frost-ed25519

Fix incorrect shares generated in roud2, step3.

### frost-secp256k1-tr

* Fix incorrect proof_of_knowledge for participant 1.
* Add missing data for participants 2 and 3.

If I have misunderstood these test vector modifications, please let me know.


conradoplg (approved):
Weird that these were wrong :thinking: 

But this looks correct, thank you!

PR #878, opened 2025-02-19
feat(frost-core): add (de)serialization for `VerifiableSecretSharingSharingSharingCommitment`

### LINKED ISSUE #870: Add a method to serialize the whole `VerifiableSecretSharingSharingCommitment<C>`.
Currently it is only possible to get `Vec<Vec<u8>>`, although `CoefficientCommitment` will be serialized in `<C::Group as Group>::Serialization`. Implementing method that provides `Vec<u8>` should not be problem since `Serialization` has fixed length.
https://github.com/ZcashFoundation/frost/blob/4d3d8822f894890e11d51b5ae4aed55cd3e50394/frost-core/src/keys.rs#L322-L328

Something like this. I'll probably open PR later.
```rust
#[cfg(feature = "codec")]
impl<C> parity_scale_codec::Encode for VerifiableSecretSharingCommitment<C>
where
    C: Ciphersuite,
{
    fn encode(&self) -> Vec<u8> {
        let tmp = self
            .serialize()
            .expect("Could not serialize `VerifiableSecretSharingCommitment<C>`")
            .concat();
        let compact_len = parity_scale_codec::Compact(tmp.len() as u32);

        let mut output = Vec::with_capacity(compact_len.size_hint() + tmp.len());

        compact_len.encode_to(&mut output);
        output.extend(tmp);

        output
    }
}

#[cfg(feature = "codec")]
impl<C> parity_scale_codec::Decode for VerifiableSecretSharingCommitment<C>
where
    C: Ciphersuite,
{
    fn decode<I: parity_scale_codec::Input>(
        input: &mut I,
    ) -> Result<Self, parity_scale_codec::Error> {
        let input: Vec<u8> = parity_scale_codec::Decode::decode(input)?;

        let generator = <C::Group>::generator();
        let serialization =
            <C::Group>::serialize(&generator).expect("serializing the generator always works");
        let serialization_len = serialization.as_ref().len();

        const ERROR_MSG: &str = "Could not decode `VerifiableSecretSharingCommitment<C>`";

        let chunks_exact = input.chunks_exact(serialization_len);

        if !chunks_exact.remainder().is_empty() {
            return Err(ERROR_MSG.into());
        }

        let serialized_coefficient_commitments: Vec<Vec<u8>> =
            chunks_exact.map(Vec::from).collect();

        Self::deserialize(serialized_coefficient_commitments).map_err(|_| ERROR_MSG.into())
    }
}
```

StackOverflowExcept1on:
Resolves #870
@conradoplg

### DIFF
--- a/frost-core/src/keys.rs
+++ b/frost-core/src/keys.rs
@@ -347,7 +347,8 @@ where
         Ok(Self::new(coefficient_commitments))
     }
 
-    /// Deserialize a whole commitment vector from a single byte vector as returned by [`serialize_whole()`].
+    /// Deserialize a whole commitment vector from a single byte vector as returned by
+    /// [`VerifiableSecretSharingCommitment::serialize_whole()`].
     pub fn deserialize_whole(bytes: &[u8]) -> Result<Self, Error<C>> {
         // Get size from the size of the generator
         let generator = <C::Group>::generator();
--- a/frost-core/src/keys.rs
+++ b/frost-core/src/keys.rs
@@ -327,6 +327,11 @@ where
             .collect::<Result<_, Error<C>>>()
     }
 
+    /// Serialize to bytes
+    pub fn serialize_whole(&self) -> Result<Vec<u8>, Error<C>> {
+        self.serialize().map(|v| v.concat())
+    }
+
     /// Returns VerifiableSecretSharingCommitment from an iterator of serialized
     /// CoefficientCommitments (e.g. a [`Vec<Vec<u8>>`]).
     pub fn deserialize<I, V>(serialized_coefficient_commitments: I) -> Result<Self, Error<C>>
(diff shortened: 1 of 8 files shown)


frost-core/src/keys.rs
```diff
@@ -327,6 +327,11 @@ where
             .collect::<Result<_, Error<C>>>()
     }
 
+    /// Serialize to bytes
```
conradoplg, line 330:
```suggestion
    /// Serialize the whole commitment vector as a single byte vector.
```

frost-core/src/keys.rs
```diff
@@ -342,6 +347,24 @@ where
         Ok(Self::new(coefficient_commitments))
     }
 
+    /// Deserialize from bytes
```
conradoplg, line 350:
```suggestion
    /// Deserialize a whole commitment vector from a single byte vector as returned by [`serialize_whole()`].
```

conradoplg (changes requested):
Looks good, would you mind adding tests? Otherwise we can add them later.

StackOverflowExcept1on:
I'll add tests and maybe update the changelog in the next 3 days.

StackOverflowExcept1on:
@conradoplg I added tests.

conradoplg (approved):
Thank you!

PR #879, opened 2025-02-24
chore(deps): avoid duplicate crates

rex4539:


frost-core/Cargo.toml
```diff
@@ -15,25 +15,25 @@ features = ["serde"]
 rustdoc-args = ["--cfg", "docsrs"]
 
 [dependencies]
-byteorder = { version = "1.4", default-features = false }
+byteorder = { workspace = true }
```
StackOverflowExcept1on, line 18: good PR, but I think we have to decide whether we want to use the syntax `byteorder.workspace = true` or `byteorder = { workspace = true }`

rex4539:
Closing due to no interest.

PR #884, opened 2025-03-20
Add post_generate to Ciphersuite trait

### LINKED ISSUE #844: Add post_generate to Ciphersuite trait
The new Ciphersuite trait methods introduced with the Taproot ciphersuite could allow making the RedPallas ciphersuite (from reddsa crate) easier to use by ensuring even-Y pubkeys using e.g. `post_dkg()`. (Currently the user needs to do it themselves if they use it with  `frost-core`, rather than the non-generic functions in `reddsa`). However, unfortunately, it's not possible to run code after trusted dealer key generation.

Add a new Ciphersuite method, e.g. `post_generate` or `post_split` (since it should be effectively be called at the end of `split()`), which will then allow the RedPallas simplification.

VolodymyrBg:
This commit adds a new `post_generate` method to the Ciphersuite trait, similar to the existing `post_dkg` method. It allows ciphersuites to customize the key generation process by post-processing the output of key generation.

This is especially useful for the RedPallas ciphersuite, which needs to perform special processing on keys to enable some optimizations. The method is integrated into the `TryFrom<SecretShare>` implementation for `KeyPackage`, which ensures it will be called when users convert their secret shares into key packages.

Fixes #844

### DIFF
--- a/frost-core/src/keys.rs
+++ b/frost-core/src/keys.rs
@@ -548,6 +548,7 @@ pub fn split<C: Ciphersuite, R: RngCore + CryptoRng>(
         }
     };
     let mut verifying_shares: BTreeMap<Identifier<C>, VerifyingShare<C>> = BTreeMap::new();
+    
     let mut secret_shares_by_id: BTreeMap<Identifier<C>, SecretShare<C>> = BTreeMap::new();
 
     for secret_share in secret_shares {
--- a/frost-core/src/keys.rs
+++ b/frost-core/src/keys.rs
@@ -564,7 +564,7 @@ pub fn split<C: Ciphersuite, R: RngCore + CryptoRng>(
     };
 
     // Apply post-processing
-    let (processed_secret_shares, processed_public_key_package) = 
+    let (processed_secret_shares, processed_public_key_package) =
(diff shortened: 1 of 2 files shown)


conradoplg:
Thank you!

I think I'd prefer to call `post_generate()` at the end of `split()` and thus it should take `BTreeMap<Identifier, SecretShare>, PublicKeyPackage`. This would ensure that the `SecretShare` will be correct without requiring receiving participants to handle it correctly.

Let me know if you're willing to do that change, otherwise we'll handle it.

VolodymyrBg:
> Thank you!
> 
> I think I'd prefer to call `post_generate()` at the end of `split()` and thus it should take `BTreeMap<Identifier, SecretShare>, PublicKeyPackage`. This would ensure that the `SecretShare` will be correct without requiring receiving participants to handle it correctly.
> 
> Let me know if you're willing to do that change, otherwise we'll handle it.

I've made some corrections

conradoplg (dismissed):
I made the suggested fix, let me know if that works for your use case

conradoplg:
@mergifyio requeue

conradoplg (approved):

PR #887, opened 2025-04-05
fix Update README.md

Hack666r:
fix old link to a new one


mpguerra:
We're no longer accepting these kinds of PRs

PR #899, opened 2025-06-05
chore(frost-core): import `String` from `alloc`

StackOverflowExcept1on:
my `#![no_std]` fork can no longer pass CI because of this:
```
    Compiling frost-core v2.1.0 (/home/runner/work/frost/frost/frost-core)
 error[E0412]: cannot find type `String` in this scope
   --> frost-core/src/tests/vectors_dkg.rs:156:18
    |
156 |     sender_num: &String,
    |                  ^^^^^^
```

@conradoplg


conradoplg (approved):
Thanks!

PR #900, opened 2025-06-06
feat: bump MSRV to Rust 1.73

### LINKED ISSUE #768: Bump MSRV to 1.81

StackOverflowExcept1on:
Resolves #768
@conradoplg Let's start at least with this change. I haven't bumped MSRV to Rust 1.81 yet, although we could do that to get rid of `thiserror-nostd-notrait` in favor of `thiserror`. I'd also keep `lazy_static` since it works in `#![no_std]` crates.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,7 +19,7 @@ jobs:
       - run: cargo build
 
   build_msrv:
-    name: build with MSRV (1.66.1)
+    name: build with MSRV (1.73)
     runs-on: ubuntu-latest
 
     steps:
@@ -34,7 +34,7 @@ jobs:
       - run: cargo update -Z minimal-versions
       # Now check that `cargo build` works with respect to the oldest possible
       # deps and the stated MSRV
-      - uses: dtolnay/rust-toolchain@1.66.1
+      - uses: dtolnay/rust-toolchain@1.73
       - run: cargo build --all-features
 
   # TODO: this is filling up the disk space in CI. See if there is a way to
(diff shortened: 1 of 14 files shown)


conradoplg:
> @conradoplg What do you think about bumping MSRV to Rust 1.81 to get rid of `thiserror-nostd-notrait`? However, `frost-ed448` will still use the old `ed448-goldilocks` dependency instead of `ed448-goldilocks-plus`?

Yes, bumping to 1.81 would be fine.

I think `ed448` was just merged into `RustCrypto` so I'd like to use that when they do a release https://github.com/RustCrypto/elliptic-curves/tree/master/ed448-goldilocks

PR #904, opened 2025-06-19
Update Zebra audit documentation links in frost-dependencies

GuroChil:
Updates links in frost-dependencies-for-audit.md to point to the correct public repository paths for redjubjub and reddsa libraries. The previous links were pointing to a private blob (zebra-private), which caused accessibility issues. This change redirects them to the public main branch, ensuring all users can access the audit documentation.


mpguerra:
We're no longer accepting these sorts of PRs: https://github.com/ZcashFoundation/frost/blob/main/CONTRIBUTING.md#pull-requests

Issue #867, opened 2025-02-07
Add min_signers to PublicKeyPackage


conradoplg:
James Katz pointed out in Discord that PublicKeyPackage does not have a min_signers. This prevents a non-Participant Coordinator from succesfully running the protocol since they will have no idea how many participants need to be involved in a signing session.

Add a min_signers field to the PublicKeyPackage.

This is a breaking change so it will need to be included only in 3.0.0

We must be careful to not break serialization compatibility if possible (we should not break that even in major versions if possible). That might be achievable using an `Option`. If it's not possible, that will require bumping the serialization version which will be a pain.

StackOverflowExcept1on:
It would also be nice to add getters `fn max_signers(&self) -> u16` and `fn min_signers(&self) -> u16` for `PublicKeyPackage`. `max_signers` seems like it might already be known from `self.verifying_shares.len()`. I also ran into this in implementation of the ROAST protocol: https://github.com/StackOverflowExcept1on/roast/blob/70dcfee21fe45979b361df107ea81cf7290c0cc7/roast-core/src/coordinator.rs#L60-L66

PR #912, opened 2025-08-26
core: add min_signers to PublicKeyPackage

conradoplg:
Closes https://github.com/ZcashFoundation/frost/issues/867

Draft for now because we want to make a release first

### DIFF
--- a/frost-core/src/serialization.rs
+++ b/frost-core/src/serialization.rs
@@ -266,9 +266,7 @@ where
 
         // The following are copied from the `serde::Deserialize` derive, and
         // are required to support `visit_map()` which in turn is required for
-        // `serde_json`. We don't need to change any of it, because of
-        // `#[serde(default)]` on the `min_signers` field which makes it work
-        // with JSON and other formats that support defaults.
+        // `serde_json`.
 
         enum Field {
             Field0,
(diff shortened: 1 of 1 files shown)


frost-core/src/tests/refresh.rs
```diff
@@ -455,7 +437,7 @@ where
 
     // Map of the verifying key of each participant.
     // Used by the signing test that follows.
```
natalieesk, line 439: Same here

frost-core/src/serialization.rs
```diff
@@ -244,3 +250,239 @@ impl<T: for<'de> serde::Deserialize<'de>, C: Ciphersuite> Deserialize<C> for T {
...
+                    })?;
+                let verifying_key = seq.next_element::<VerifyingKey<C>>()?.ok_or_else(|| {
+                    serde::de::Error::invalid_length(
+                        2usize,
+                        &"struct PublicKeyPackage with 4 elements",
+                    )
+                })?;
+
+                // For the `min_signers` field, fill it with None if
+                // `next_element()` fails (i.e. there are not other elements)
```
natalieesk, line 389: Typo: "there are not other elements" change to "there are no other elements"

PR #971, opened 2025-12-18
core: fix unused warning due to compiler bug

conradoplg:
Anything that uses `#[zeroize(skip)]` gives an unused warning, but that is a bug https://github.com/rust-lang/rust/issues/147648

So just silence the warning for now

### DIFF
--- a/frost-core/src/keys.rs
+++ b/frost-core/src/keys.rs
@@ -1,5 +1,7 @@
 //! FROST keys, keygen, key shares
 #![allow(clippy::type_complexity)]
+// Remove after https://github.com/rust-lang/rust/issues/147648 is fixed
+#![allow(unused_assignments)]
 
 use core::iter;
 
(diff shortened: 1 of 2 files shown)


frost-core/src/keys.rs
```diff
@@ -1,5 +1,7 @@
 //! FROST keys, keygen, key shares
 #![allow(clippy::type_complexity)]
+// Remove after https://github.com/rust-lang/rust/issues/147648 is fixed
+#![allow(unused_assignments)]
```
adr1anh:
```suggestion
#![expect(unused_assignments)]
```
Using `expect` will automatically trigger a warning when the compiler is fixed

conradoplg: TIL, thanks! However, that will emit warnings for whoever is using the crate in pre-1.92 Rust :frowning_face:

PR #1008, opened 2026-01-28
fix issues from specific feature combinations

conradoplg:
Most of them are due to serialization roundtrips being testes but that requires the `serialization` feature so it would fail to compile without it. I just gated the tests under it.


natalieesk (approved):
Looks good

PR #1040, opened 2026-03-30
core: add ZeroizeOnDrop for dkg::round2::Package

### LINKED ISSUE #1018: Implement Zeroize for round1::Package<C> and round2::Package<C> in DKG module
PR #945 added Zeroize and ZeroizeOnDrop derives to key-related structs including SecretShare, KeyPackage, and DKG SecretPackage types. However, the DKG round Package types were not included.

Issue
Both round1::Package<C> and round2::Package<C> in the frost::keys::dkg module contain sensitive cryptographic material:

round2::Package contains signing_share: SigningShare<C> — a participant's secret share
round1::Package contains proof_of_knowledge: Signature<C> — includes a secret nonce scalar
Neither type implements Zeroize or ZeroizeOnDrop. This is inconsistent with the parallel SecretPackage types (which do implement these traits after PR #945) and leaves secret material in memory when these packages are dropped.

Downstream consumers who collect received packages in a BTreeMap during DKG cannot zeroize the secret material after part3() completes, since signing_share is pub(crate) and round2::Package has no Zeroize impl.

Suggested Fix
Add Zeroize and ZeroizeOnDrop to the derive macros:

#[derive(Clone, Debug, PartialEq, Eq, Getters, Zeroize, ZeroizeOnDrop)]
pub struct Package<C: Ciphersuite> { ... }
Non-secret members like header can use #[zeroize(skip)] if needed.

Impact
Additive change only — no breaking API impact. Downstream consumers get automatic memory zeroization on drop without code changes.

Worth noting: the research found that frost-core already did a zeroization pass in PR #945 (closing issue #877), so this fits naturally as a follow-up to that work. The round1::Package angle (proof of knowledge containing a nonce scalar) broadens the request beyond just round2::Package.

conradoplg:
Might close https://github.com/ZcashFoundation/frost/issues/1018 if we agree `round1::Package` is not needed

### DIFF
--- a/frost-core/src/keys/dkg.rs
+++ b/frost-core/src/keys/dkg.rs
@@ -223,7 +223,7 @@ pub mod round2 {
     /// # Security
     ///
     /// The package must be sent on an *confidential* and *authenticated* channel.
-    #[derive(Clone, Debug, PartialEq, Eq, Getters)]
+    #[derive(Clone, Debug, PartialEq, Eq, Getters, Zeroize, ZeroizeOnDrop)]
     #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
     #[cfg_attr(feature = "serde", serde(bound = "C: Ciphersuite"))]
     #[cfg_attr(feature = "serde", serde(deny_unknown_fields))]


natalieesk (approved):
Looks good

